PR #13, opened 2026-02-25
refactor(desktop): WS event router, hook split, and stale-socket fix

doruchiulan:
## Summary

- **Split renderer hooks**: `useDaemon` → `useAppInit` (connection + data loading) and `useChat` → `useChatSession` (per-chat behaviour)
- **Add `ws-event-router`**: centralises `DaemonEvent` dispatch to Zustand stores, keeping components decoupled from WebSocket logic
- **Fix stale-socket reconnect bug**: `DaemonClient.connect()` now captures the socket in a closure-local `const`. `onclose`/`onerror` handlers guard with `if (socket !== this.ws) return` so that close events from sockets replaced by a cleanup/re-mount cycle (e.g. React Strict Mode) are silently dropped instead of logging a spurious "disconnected" and scheduling an unnecessary reconnect

## Test plan

- [x] `pnpm --filter @mainframe/desktop test` — 232 tests pass
- [x] New `client.test.ts` covers both the stale-socket scenario and normal unexpected-disconnect reconnect


doruchiulan:
retest

PR #190, opened 2026-04-08
feat: suppress mobile push when desktop is active

doruchiulan:
## Summary

- Electron main process detects OS-level idle using `powerMonitor.getSystemIdleTime()`, reports state transitions (`active`/`idle`) to daemon via `POST /api/device/activity`
- Daemon tracks `desktopActive` boolean with a 6-minute staleness timer — if the desktop crashes or quits without reporting, push resumes automatically
- `PushService.sendPush()` skips all push notifications when desktop is active
- Idle threshold: 5 minutes of no mouse/keyboard activity

## Test plan

- [x] 9 unit tests for `PushService` (suppression, staleness timer, timer reset)
- [ ] Manual: verify no mobile push while desktop is focused
- [ ] Manual: verify mobile push resumes after 5 min idle or desktop quit
- [ ] Manual: verify mobile push resumes within 6 min if desktop crashes


packages/desktop/src/main/idle-reporter.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const POLL_INTERVAL_MS = 30_000; // 30 seconds
+const IDLE_THRESHOLD_S = 5 * 60; // 5 minutes in seconds
+const DAEMON_PORT = process.env['DAEMON_PORT'] ?? '31415';
+
+let currentState: 'active' | 'idle' = 'active';
+let pollTimer: ReturnType<typeof setInterval> | null = null;
+
+function getDaemonUrl(): string {
+  return `http://127.0.0.1:${DAEMON_PORT}`;
```
doruchiulan: don't we have an env variable for daemon host ?

PR #267, opened 2026-04-30
fix(claude/events): drop dispatch prompt when CLI normalizes it to a text block

doruchiulan:
## Summary

PR #264 caught the subagent dispatch prompt only on the **string-content** path. Live testing showed the duplicate pill still appearing. Tracing it: the CLI's `normalizeMessages` (`claude-code/src/utils/messages.ts:782-793`) wraps any string-content user message into `[{type: 'text', text: <string>}]` **before** yielding to stream-json:

```ts
case 'user': {
  if (typeof message.message.content === 'string') {
    ...
    content: [{ type: 'text', text: message.message.content }],
    ...
```

So the daemon receives the dispatch prompt as a single text block — the previous string-only guard never fires. The event then falls through to the array-content text-block branch and `onCliMessage` produces the system pill the screenshot showed.

Repro: session `22ebe864-…` in `qlan-home-hub` — three "Run \`echo hi\` via Bash and report the output." pills next to the three `general-purpose` Task cards.

## Fix

Add a `parent_tool_use_id != null` guard inside the array-content text-block branch, alongside the existing `isLocalCommandWrapper` and `isInterruptMarker` suppressions. Skill injections still take the `isSkillInjection` branch above and surface as before; only the duplicate dispatch prompt is dropped.

## Test plan
- [x] `pnpm --filter @qlan-ro/mainframe-core test src/__tests__/claude-events.test.ts` — 25/25 pass; new regression covers the array-of-text-blocks shape produced by `normalizeMessages`.
- [x] `pnpm --filter @qlan-ro/mainframe-core build` — clean.
- [ ] Manual: dispatch ≥1 Agent in a live chat (e.g. session `22ebe864-…` in `qlan-home-hub`). Verify no duplicate prompt pill next to the Task card; subagent skill loads and tool_use children still surface.

## Related
- PR #264 — same bug, string-content path only (the path that doesn't fire in practice because of `normalizeMessages`)
- PR #259, PR #261 — earlier surfaces of the same family of subagent → parent-thread leak

### DIFF
--- a/.changeset/metal-friends-invite.md
+++ b/.changeset/metal-friends-invite.md
@@ -1,0 +1,5 @@
+---
+'@qlan-ro/mainframe-core': patch
+---
+
+Drop the duplicate subagent dispatch prompt when it arrives in the array-of-text-blocks shape. PR #264 caught the string-content path, but the CLI's `normalizeMessages` (`utils/messages.ts:782-793`) wraps any string-content user message into `[{type: 'text', text: <string>}]` before yielding to stream-json — so in practice the daemon receives the prompt as a single text block, not as a string. The new guard fires inside the array-content text-block branch, after the existing skill-injection check, so subagent skill loads still surface and only the duplicate prompt is suppressed.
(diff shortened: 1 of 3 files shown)


doruchiulan:
Superseded by #268. That PR takes the structural fix instead of suppression — every subagent event with `parent_tool_use_id` is routed through `onSubagentChild`, tagged with `parentToolUseId`, and inlined inside the parent Task card by `groupTaskChildren`. The duplicate dispatch prompt is no longer dropped; it surfaces inside the Task card body alongside the rest of the subagent activity. Closing in favor of the architectural approach.

PR #408, opened 2026-07-07
feat(cli): add `mainframe --version` and show version in status/health

doruchiulan:
Surfaces the daemon's own version — there was no version plumbing anywhere before (no constant, no `/health` field, no build injection).

## What's new
- **`mainframe --version`** (also `-v` / `version`) → prints `mainframe <version>` for the installed binary.
- **`mainframe status`** → now shows a `Version:` line (the **running** daemon, read from `/health`).
- **`GET /health`** → returns a `version` field.

## How the version is resolved
Single source of truth in `packages/core/src/version.ts`. The daemon is bundled to a single `daemon.cjs` that ships no `package.json`, so the version is inlined at build time via esbuild `define: { __DAEMON_VERSION__ }` in **both** bundlers (Tauri sidecar + standalone). Dev (`tsx`) and unbundled `node dist/index.js` runs fall back to reading `package.json`.

## The `--version` fast path
`--version` is handled by a first-import guard (`cli/early-flags.ts`) so it answers **before** the logger (pino file destination), DB, or server are evaluated. Without this, an immediate `process.exit(0)` right after boot leaks a `logger initialized` JSON line to stdout and trips pino's `sonic boom is not ready yet` flush race. Now it's a clean `mainframe <version>`, exit 0.

## Verification
- Unit test (`version.test.ts`) — `DAEMON_VERSION` matches `package.json` on the fallback path.
- `mainframe --version` / `-v` / `version`: clean output, exit 0, on both the compiled `dist` (fallback) **and** a real esbuild bundle (the `define` path — literal inlined, `__DAEMON_VERSION__` fully replaced).
- Booted a daemon: `/health` → `"version":"2.0.0-rc.1"`, `status` → `Version: 2.0.0-rc.1`.
- Typecheck clean; no test pins the `/health` shape.

### DIFF
--- a/.changeset/daemon-version-command.md
+++ b/.changeset/daemon-version-command.md
@@ -1,0 +1,9 @@
+---
+"@qlan-ro/mainframe-core": minor
+---
+
+Surface the daemon version. `mainframe --version` (also `-v` / `version`) prints
+the installed binary's version, `mainframe status` shows the **running** daemon's
+version, and `GET /health` now returns a `version` field. The version is inlined
+into the bundle at build time (esbuild `define`), with a `package.json` fallback
+for dev and unbundled runs.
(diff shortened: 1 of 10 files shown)


doruchiulan:
retest this please

PR #512, opened 2026-07-24
fix(ui): dismiss the image lightbox by clicking the image or dimmed area

doruchiulan:
Closes #276 (todos plugin).

## Summary

The lightbox only dismissed when you clicked the empty bands above or below the image. Radix's "interact outside Content" detection fires for pointer-downs on the overlay, and `DialogContent` wrapped the image tightly — so a click on the image itself counted as *inside* Content and did nothing. Users had to aim at the letterbox.

Both `ImageLightbox` (multi-image) and `ZoomableImage` (single) had the same shape, so the fix landed in a new shared `LightboxSurface` they both render. It owns the dialog content, the image, and one click handler:

```ts
if (event.target === event.currentTarget || event.target === imageRef.current) onDismiss();
```

Dismiss fires for the content box and the image; anything else — nav chrome, counter, close button, all absolutely positioned over the surface by the caller — is left alone. That is an identity check rather than `stopPropagation` on every control, so a control added later can't silently start dismissing.

Escape and the close "X" are unchanged.

## Route

`route:no-spec`. The brief was the spec; no product decisions were open.

## Review

`reviewer` and `quality-reviewer` both ran on the diff. Findings were nits only (naming and comment precision in `LightboxSurface`), fixed in `67be7427`. Nothing was skipped.

## QA

Autonomous run, 12 scenarios, all pass. jsdom `fireEvent` dispatches to whatever node you name regardless of visual overlap, so it cannot answer the question this fix actually turns on — whether nav chrome layered over the dismiss surface receives the real click. QA mounted both components in a throwaway Vite entry and drove them with real coordinate mouse events:

| Scenario | Result |
|---|---|
| Backdrop click dismisses (both components) | pass |
| Image click dismisses (both components) | pass |
| Next / Prev navigate, do not dismiss | pass |
| Counter click is a no-op | pass |
| Close button dismisses | pass |
| Escape dismisses (both components) | pass |
| Unit suite (22 tests) + UI typecheck | pass |

Not verified: the five call sites that open the lightbox (`InlineImageThumbs`, `AssistantMessage`, `TaskAttachments`, `SessionAttachmentsGrid`, `ImageViewer`) still pass props correctly — QA drove `LightboxSurface`'s two direct callers rather than seeding real image attachments. The diff is confined to those two callers and the shared surface.

**Correction to the brief:** it assumed clicking the image did not dismiss *and should*. That reads as if image-click were the only gap — in fact both the image and the side bands were dead, and both now work. The brief's "zoom/pan interactions must not be swallowed" is N/A: there is no pan/zoom gesture logic in this component (it is click-to-open-full-size only).

### DIFF
--- a/.changeset/lightbox-click-to-dismiss.md
+++ b/.changeset/lightbox-click-to-dismiss.md
@@ -1,0 +1,7 @@
+---
+'@qlan-ro/mainframe-ui': patch
+---
+
+Dismiss the image lightbox by clicking the image or the dimmed area around it.
+
+The lightbox closed only on clicks that landed on the overlay — the empty bands above and below the image — because the image and the dimmed space beside it sit inside the dialog's content box. Clicking anywhere that is not a control now closes it; the prev/next buttons, the counter, Escape, and the close button behave as before.
--- a/packages/ui/src/features/chat/parts/ImageLightbox.tsx
+++ b/packages/ui/src/features/chat/parts/ImageLightbox.tsx
@@ -6,14 +6,15 @@
  * The parent owns the open index (`null` = closed); this renders the current
  * image plus prev/next controls, a counter, and ArrowLeft/ArrowRight keyboard
  * nav (wrapping at the ends). Restores the desktop multi-image gallery
- * affordance that the single-image `ZoomableImage` didn't cover. Built on our
- * shadcn `Dialog` so no new dependency.
+ * affordance that the single-image `ZoomableImage` didn't cover. The dialog
+ * shell, image, and click-to-dismiss behavior live in `LightboxSurface`.
  *
  * For a single image the nav chrome is omitted (it degrades to a plain zoom).
  */
 import { useEffect } from 'react';
 import { ChevronLeft, ChevronRight } from 'lucide-react';
-import { Dialog, DialogContent, DialogTitle } from '@/components/ui/dialog';
+import { Dialog } from '@/components/ui/dialog';
+import { LightboxSurface } from './LightboxSurface';
 
 export interface LightboxImage {
   src: string;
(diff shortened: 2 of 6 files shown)


doruchiulan:
## Live QA — PASS (7/7)

Ran against the branch in a throwaway worktree (Rust daemon 32313, UI 6064), headed browser, real mouse clicks for the coordinate-sensitive cases. Fixtures: three PNGs of distinct sizes (120×80, 200×100, 60×160) attached to one message so the image index is unambiguous.

| # | Check | Result |
|---|---|---|
| S1 | Click the image itself dismisses (multi-image `ImageLightbox`) | **PASS** — opened thumb 0 → `{current: 120x80, counter: "1 / 3"}`; click on `image-lightbox-current` → dialog gone, 0 overlays |
| S2a | Click the dimmed band *inside* the content box dismisses | **PASS** — box spans x60–1440/y435–515 while the image occupies only x690–810; real click at (200,475) closed it |
| S2b | Click the scrim *outside* the box dismisses | **PASS** — click at (750,150) closed it |
| S3 | Nav chrome must **not** dismiss | **PASS** — next → `200x100 "2 / 3"`, next → `60x160 "3 / 3"`, click on the counter → unchanged and still open, prev → `200x100 "2 / 3"`, ArrowRight → `60x160 "3 / 3"`. Never dismissed. |
| S4 | Single-image `ZoomableImage` variant | **PASS** — reached via the Files viewer; `chat-image-zoom-dialog` opened at 300×200, image click closed it; reopened, band click at (200,475) closed it |
| S5 | Escape closes | **PASS** — dialog count 0, body `pointer-events` lock released |
| S6 | Built-in Close button closes | **PASS** — a control click that *should* dismiss still does |

Surface check on the Files viewer used for S4: `viewer-shell-status` read `PNG · 300×200`, zoom-in worked. No visual glitches.

### Not covered (pre-existing, not this PR)
The composer's "Open file…" fuzzy search can't reach images at all — `add_file_hit` in `packages/core-rs/crates/mainframe-server/src/routes/files.rs:225` returns early for `has_binary_extension(rel)`, so PNGs never appear in results (a sibling `.ts` file in the same fixture repo does). Present on `main`; I reached `ImageViewer` through the Recent list instead. Worth a separate issue if images are meant to be openable that way.

No defects found in the change itself.

PR #520, opened 2026-07-25
feat: offer to move a session into a worktree the agent created (#269)

doruchiulan:
Closes #269.

When an agent creates a worktree mid-session, the session stays where it was — the user has to notice, then move manually. This surfaces the new worktree as an offer in the composer: accept and the session rebinds and restarts there, dismiss and it stays dismissed for that chat. Sessions already isolated in a worktree get a "Move to another worktree" list in the existing popover.

## How it's built

A pure eligibility scan decides what counts as an offer (6 gates, no IO, no clock, no locks). A registry holds per-chat offer state and coalesces rescan bursts into one trailing pass. The event sink watches for worktree-shaped tool uses and fires a rescan once per successful tool-result batch — so the offer appears right after `git worktree add` returns, not on a timer.

Offer state rides on `ChatThreadState` alongside `backgroundTasks`, not on `Chat`, whose `sameComposerConfig` allowlist would swallow it. The subscribe frame carries a full offer snapshot, sent unconditionally so an empty list clears stale offers on reconnect — it is the only re-seed path, since offers never appear on the `Chat` payload.

## Artifacts

- Spec: `docs/specs/2026-07-25-todo-269-worktree-switch-offer.md`
- Plan: `docs/plans/2026-07-25-todo-269-worktree-switch-offer-plan.md`

## Decisions a reviewer should second-guess

- **The switch guard lives only in the registry** (`ChatOffers.switching`), not in `Guards`. `join_flight` exists to *await* an in-flight operation; the spec requires accept to fail 409 rather than wait, so there is no awaiter.
- **Accept returns 409 rather than queueing.** The plan text said queue; the implementation returns 409 and the banner disables every other row while a switch restarts, so the common double-click case is mostly unreachable. When 409 does fire, the row returns to actionable and the server message surfaces in a toast. The daemon owns offer removal — nothing is dropped optimistically.
- **Canonicalization is rescan-only.** All four path sources are canonicalized when the offer is minted; the sync entry points compare verbatim, because callers hand back the canonical path the rescan produced. Pinned by a test asserting a raw non-existent path returns `NotPending`.
- **`dismiss` validates and removes under one lock**, not the plan's two acquisitions, which would let two concurrent dismisses both pass validation.
- **The shared body validator returns `Option`, not `Result<_, Response>`.** Carrying a ready-made 400 in the error variant trips `clippy::result_large_err` under this workspace's lint set.
- **A real fix rode along:** `routes/worktree.rs` stripped `refs/heads/` with `.replace()`, which mangles a branch named `feat/refs/heads/x`. It now uses the extracted `short_branch` helper's `strip_prefix`.

## Known gaps

- **Two pre-existing 300-line violations grew.** `chat-thread-state.ts` 443 → 518 and `chat-thread-controller.ts` 331 → 370. Both were over the limit before this work and the plan directed the additions into them; splitting either is a refactor with its own blast radius. Flagged rather than silently folded in.
- **Two new Rust test files exceed the cap** (`worktree_offer/tests.rs` 598, `worktree_offer_scan.rs` 331 including 269 test lines). All three production files are well under (210, 182, 62).
- **A detached worktree attached from the move list records `branchName: 'detached'`** — a literal string standing in for a real state. Consistent with the pre-existing Existing-tab path; deliberately not changed, since it would alter a shipped surface's persisted data for a case the spec didn't scope.
- **One banner test runs on real timers** with a why-comment: `userEvent`'s delay loop deadlocks against the fake clock. It asserts click routing; the 2s self-clear has its own fake-timer test.

## Not run

The reviewer/quality-reviewer panel and the QA smoke test were skipped — this session ran out of token budget. Worth running before merge rather than treating the green checks as sufficient.

## Verification

```
cargo fmt --check                          EXIT=0
cargo clippy --all-targets -- -D warnings  EXIT=0
cargo test -p mainframe-chat               151 passed, 0 failed
cargo test -p mainframe-server             180 lib + all integration suites, 0 failed
packages/types  tsc --noEmit               EXIT=0
packages/types  vitest run                 14 files / 112 tests passed
packages/ui     tsc --noEmit               EXIT=0
packages/ui     vitest (4 files, singly)   60 passed, 0 failed
```

Every test was written red-first and proven to fail before its implementation landed.


doruchiulan:
## Live QA — worktree switch offer

Manual run against the branch (Rust daemon on a throwaway fixture repo, `~/Projects/qa520-repo`, six worktrees created by real Claude turns). Watch mode, UI + API + filesystem assertions.

| # | Scenario | Status |
|---|---|---|
| S1 | First worktree a session creates raises an offer | **FAIL** |
| S2 | Accept rebinds the session to the worktree | PASS |
| S3 | Dismiss persists (no re-offer after a later trigger, survives reload) | PASS |
| S4 | Eligibility — another chat's bound worktree is never offered | PASS |

### S1 FAIL — the first worktree a session creates is never offered

Reproduced 2/2 in two independent chats:

- **Chat A** — turn 1 created `qa520-wt-a` → **no banner** (polled 20×6s, plus a reload to force a fresh subscribe snapshot). Turns 2 and 3 created `wt-b` and `wt-c` → both offered immediately.
- **Chat C** (genuine first turn of a fresh chat) — created `qa520-wt-d` → **no banner** across 16 polls (~96s). Its second trigger created `wt-e` → offered by poll [2].

Each time the worktree really was created (`git worktree list`) and the turn really ran (a Bash `tool_use` is present in the message payload), so the trigger side is fine — the offer just never comes.

This matches `baseline_and_pending` in `worktree_offer/rescan.rs` taking its `None` branch: with no baseline present, the scan *seeds* the baseline from the current listing, absorbing the just-created worktree instead of raising an offer. The spec's own defensive rule says that path should be unreachable, because the baseline is captured **eagerly at chat activation, before the adapter spawns**:

> **Baseline** — the set of registered worktree paths, captured eagerly at chat activation (before the adapter spawns), so no trigger can precede it.

`ChatManager::seed_worktree_baseline` does override the trait default, and `lifecycle_manager.rs` awaits it ahead of the early returns — so the seeding call exists; something about its ordering or its per-chat keying isn't holding for the first trigger. Worth a test that asserts an offer on the **first** worktree of a brand-new chat, since every existing repro passes on the second.

Impact: this breaks AC #1 and the feature's headline use case — the common shape is one worktree per session, which is exactly the case that never prompts.

### S2 PASS — accept rebinds

Chat C went `worktreePath=None` → `/Users/doruchiulan/Projects/qa520-wt-e`, `branchName=qa520-e`. The Settled state rendered verbatim ("Session is now in … on qa520-e.") and self-removed on the next poll; the binding survived a reload; the Claude session jsonl relocated to `-Users-doruchiulan-Projects-qa520-wt-e/`; and the resumed agent recalled both earlier commands and reported its cwd as the worktree — so the `--resume` handoff kept history.

### S3 PASS — dismiss persists

Dismissing `wt-b` from a 2-row multi-offer collapsed the banner to the single-offer form for `wt-c`. A later trigger creating `wt-f` produced `[wt-c, wt-d, wt-f]` — `wt-b` never returned — and the same three rows survived a page reload, so the dismissal is carried in the subscribe snapshot, not just client state.

### S4 PASS — eligibility

With chat C bound to `qa520-wt-e`, chat A's offers never included it, while the unbound `qa520-wt-d` did appear. The "not another chat's current worktreePath" gate holds.

### Cosmetic — stale `sessionFilePath` after an accepted switch

After the rebind, the chat row's `sessionFilePath` still points into the pre-switch directory, and that file no longer exists there. Nothing breaks today: `is_claude_transcript_present` falls back to the path derived from the project path, which resolves inside the worktree. But the comment at `adapter-claude/src/transcript.rs:46` calls the stored path "authoritative — survives worktree moves", which is no longer true — either update the stored path on accept, or fix the comment.

### Not covered

Multi-project offers and the `EnterWorktree` trigger variant (only the Bash-command trigger was exercised).

_Fixture repo, worktrees, chats and project all removed after the run._

doruchiulan:
image

PR #521, opened 2026-07-25
feat: Setup Advisor — recommend the automations a project has earned (#191)

doruchiulan:
Closes #191.

Mainframe knows what a project is made of the moment you open it — TypeScript, Prisma, a Supabase client, a `.github/workflows` directory — and until now it did nothing with that. The Setup Advisor reads those signals and turns them into a short list of MCP servers, skills, hooks, subagents, and plugins the project has actually earned, each one a click away from the clipboard.

**It never writes to your project.** Every recommendation is a command or a file body you copy and run yourself. The sheet says so per tab, in the language that fits what that tab contains: a tab of shell commands promises "nothing runs until you paste it in your terminal", a tab of file bodies promises "nothing is written until you paste it into your project".

## What's here

- **Fingerprinting** (`packages/core-rs/crates/mainframe-server/src/setup_advisor/`) — manifests, lockfiles, config files, and directory shape across 11 ecosystems, bounded by a 15-second budget and a file cap.
- **A 76-rule catalog** across five families, each rule stating the signal that fired it.
- **`GET /api/projects/:id/automation-recommendations`** — returns the fingerprint plus the ranked recommendations, or 504 when the scan outruns its budget.
- **The sheet** (`packages/ui/src/features/setup-advisor/`) — one tab per category, an evidence disclosure listing what was detected, and a copy button per row.

## Decisions

**Aggregator rules are allowed, and they say so.** Per the user's ruling on the brief, recommendations sourced from community aggregators carry their repo and install count in a distinct amber chip; vendor-official rules show plain attribution and first-party rules show none. An aggregator install stays a decision, not a surprise.

**No auto-apply, no "apply all".** The spec considered writing files directly. Copy-only means a user who doesn't understand a recommendation can't be harmed by accepting it.

## Review findings addressed

The reviewers found 12 must-fix items; all 12 are in `477f568c` and `1b47a7c6`. The load-bearing ones:

- **Scan bounds.** A symlink cycle could walk forever; a huge manifest could be read whole into memory. Now: an `entered: HashSet<PathBuf>` guard, a manifest size cap, and a wall-clock budget that returns 504 rather than hanging the request.
- **The payload lied about itself.** Rows whose payload is a file body were previewed as `command.split('\n')[0]`, so a JSON hook rendered as a bare `{` and a markdown scaffold as `---`. Payload kind is now derived from `targetPath` and drives both the preview and the footer copy.
- **A stale report could appear under a new project's name.** The clearing effect runs after commit, so the render that first saw the new `projectId` still held the old report. Gated at the prop boundary on the id the report was fetched for.
- **`-y` stripped from third-party install commands** — auto-confirming a package install from an aggregator is exactly the decision the provenance chip exists to preserve.

## Deliberately not done

- **The `-g` flag on the 35 skills rules.** Filed as a security finding, but dropping it changes whether skills land globally or per-project — a product decision, not a fix. Left as-is.
- **The `adapters` field on TS-side rules** stays pinned; moving it is blocked by the TypeScript-version embargo.
- **`aggregator_tooling.rs` keeps its own evidence strings** rather than reusing `families::RUFF_CONFIG`/`TSCONFIG`. Those constants carry a leading `"a "`, so reusing them would have silently reworded two rules' user-visible text.
- **`build_report`'s `None` → 504 line is untested** — reachable only by a real timeout, which the QA run exercised end-to-end but no unit test covers.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` all exit 0 — 2258 tests passing. UI typecheck and the touched vitest files pass.

QA smoke-tested 8 scenarios against a live daemon: a real project fingerprint (200 in 0.23s, 8 recommendations), the read-only promise per tab *and* on disk (`git status` byte-identical before and after, nothing under `.claude/` touched), full-command clipboard copy, provenance rendering, project switching, an empty project, a real 15s timeout to a 504 and the UI error state, and daemon-down plus route-500 retry recovery. Zero unexpected console errors.

## Repo defect found during QA — not fixed here

`.agents/launch-test-browser.sh` boots `@qlan-ro/mainframe-core`, the Node daemon retired in #510, so every API route under test 404s. It also runs an unconditional `pnpm install`, which in a worktree strips all 6971 `packages/mobile` lines from `pnpm-lock.yaml` (the submodule isn't populated there). Both are unrelated to this PR and are going up separately.


doruchiulan:
## Live QA — 3/3 PASS

Manual QA against a running app built from this branch (Rust daemon on a throwaway port, headed browser). Two disposable fixture projects, both deleted afterwards.

| # | Scenario | Class | Mode | Status |
|---|----------|-------|------|--------|
| S1 | Rule-table fidelity on a single-signal fixture | happy | api+ui | **PASS** |
| S2 | `hasClaudeConfig` toggles the mutually-exclusive plugins pair | edge | api | **PASS** |
| S3 | No stale report after a project switch | regression | ui | **PASS** |

### S1 — recommendations match the rule table exactly

Fixture A: `package.json` depending only on `@supabase/supabase-js`, plus an `index.js`. Fingerprint came back `databases:["supabase"], frameworks:[], languages:[], gitHost:null, hasClaudeConfig:false, fileCount:2`.

Returned exactly: `mcp-supabase`, `skills-supabase`, `skills-create-migration`, `subagents-performance-analyzer`, `plugins-claude-code-setup`.

That is precisely what the catalog dictates for this fingerprint — `mcp-playwright`/`mcp-context7` stay silent on empty frameworks, `mcp-github` on a null gitHost, `skills-feature-dev` on empty languages. Nothing padded the list out to the per-category cap, which matches the "Never pads" intent.

### S2 — the claude-config pair is genuinely exclusive

Adding a `CLAUDE.md` flipped `hasClaudeConfig` to true. `plugins-claude-code-setup` disappeared and `plugins-plugin-dev` took its place, with the expected evidence string ("a .claude/ directory or CLAUDE.md at the repo root"). The two never appeared together. Every other row was byte-identical across the two runs, so the toggle moved only the pair it should.

### S3 — no stale rows under the new project

Fixture B (`Cargo.toml` only) recommends a disjoint set — `skills-rust`, `skills-project-conventions`, `subagents-test-writer`, `plugins-claude-code-setup`, `plugins-rust-analyzer-lsp` — so any leak from A would be unmistakable.

Opened the sheet under A (showed `mcp-supabase`), closed, switched the active session to B, reopened with a 50 ms DOM sampler running. **237 samples, zero frames containing a supabase row.** The sheet went from empty straight to B's rows, and the `mcp` tab was absent entirely under B — correct, since B produces no MCP recommendations.

Worth recording for whoever reads the guard later: **the switch-*while*-open race is not user-reachable.** The sheet is modal — `body { pointer-events: none }` plus a full-screen scrim above the sidebar — so a project cannot be switched while it is open. The `reportForProject === projectId` gate still earns its place on the close → switch → reopen path, which is what S3 actually exercises, but the comment above it describes a path the UI does not currently expose.

### Surface check

Copy is the sheet's central affordance and it works — the button flips to "Copied". I did not assert the pasted text; `navigator.clipboard.readText()` blocks on a permission prompt in a headed browser. The advisor also wrote nothing to either fixture directory (no `.claude/`, no config files), which matches the copy-only contract.

### No defects found.

One unrelated observation, pre-existing and **not** from this PR: the sidebar session rows use an unkeyed `data-testid="sessions-row"` for every row, so they can only be addressed positionally. The repo's own rule asks for testids keyed by domain id.

PR #522, opened 2026-07-25
Automations editor: named values, webhook secrets, and composer-parity fields (#234)

doruchiulan:
Rebuilds the automations editor around **values you can name and reuse**. Closes #234.

The plan proposed six PRs; the work landed coherently on one branch, so this is the whole feature.

## What changed

- **Every text field in an automation is one field.** `TriggerTextField` replaces the old chip fields: `$name` references, `/` skills and `@` files all fire from the same trigger engine the chat composer uses.
- **A Set value step names a result once**, and later steps type `$name`. Renaming that value rewrites every step that referred to it. Write `${name}` where a bare `$name` would run into surrounding text, as in `todo/${id}`.
- **Each step that produces a value carries its own name** (`outputName`), so reordering steps no longer silently repoints a reference at a different step. Two values sharing a name is reported as an error on **both**, rather than one quietly winning.
- **Webhook triggers register from the editor**, which now shows the signing secret alongside the URL — without it a registered hook rejected every delivery. The secret is returned once on register and revealed on request; reads are structurally incapable of returning it.
- **Problems are reported on the step that caused them**, including the ones the daemon finds at save time (previously a toast). A reference to a name nothing defines is a **warning**, not an error, so a prompt containing `$HOME` no longer blocks saving.
- **One-off (`once`) schedules**, plus the daemon/types groundwork for automation variables.

## Plan

`docs/plans/2026-07-25-todo-234-automations-editor-plan.md` (route `no-spec` — the brief is the spec, no PM stage).

## Product decisions (the user's, taken during the design pass)

1. **Automation text fields behave exactly like the real composer**, including `/` and `@`, with `$` added for variables — the trigger engine was adapted rather than a second mechanism built.
2. **`$name` is plain text, not a chip.** Fields are textareas; losslessness lives on the wire, not in the rendering.

## Planner decisions

1. Route 1 — lift the composer's trigger engine rather than re-implement it.
2. `$` namespace semantics: one flat namespace per automation.
3. Dotted digging (`$result.field`) stays supported.
4. Stale-chip language dropped from the UI copy.
5. `ChipText` stays the wire shape (mobile-co-owned contract, additive only).
6. Webhook **Register** is a real daemon route, not a client-side URL construction.
7. Structural pickers (if-condition left side, repeat list) keep the `TokenRef` model — only prose fields moved to `$name`.
8. `MoreOptions` survives for AskMe/Notify.
9. Rename-rewrite covers `set_variable` names only.

## Review verdicts

`reviewer` + `quality-reviewer` ran on `git diff main...HEAD`; all must-fix findings were routed back and fixed in `82f5ac0b` (naming model, lossless round trip, `${name}` delimited form, unresolved-name warning downgrade, webhook secret handling). A follow-up audit pass caught leftovers the reviews missed — three orphaned `TokenPicker` props and five stale `ChipField` comments — fixed in `860b9a9a`.

## QA (8 scenarios, live Rust daemon on an isolated data dir)

| # | Scenario | Result |
|---|---|---|
| S1 | Naming survives reordering (the core defect) | **PASS** — verified twice, second time on current HEAD |
| S2 | Round trip lossless | **PASS** — structural token parts survive persistence |
| S3 | `${name}` delimited form | **PASS** — braced mid-word ref persists; `todo/$agent_result` stays literal, as designed |
| S4 | Duplicate names reported | **PASS** — daemon flags **both** step ids |
| S5 | Unresolved name warns, doesn't block | **PASS** — `$HOME` warns, Save enabled |
| S6 | Webhook signing secret | **PASS** — shown once, same on re-register, absent from DOM/record/list route after reload |
| S7 | Composer parity `/` `@` `$` | **PARTIAL** → fixed in `cf4c7f56` |
| S8 | Nothing else broke | **PASS** — schedules (incl. one-off), add/delete/reorder/save/cancel |

No scenario failed. Two expectations QA couldn't express as written: "change nothing, save" (Save is correctly disabled while clean) and "chips/pills" (this rebuild deliberately replaced chips with text).

**S7's defect, fixed here:** fields belonging to no agent step took the first *installed* adapter, and only Claude answers the skills route — so with Codex installed first, `/` opened nothing in Notify, Set value and action inputs. The fallback now tries each installed adapter until one answers.

## Known gaps, not fixed here

- **Duplicate-name copy overlap.** Typing a name another value already uses is refused at the keystroke, so the step stays unnamed and its card also says "Give this value a name." Both messages are true and the pane explains the real cause. Committing the duplicate instead would make a rename round trip lossy (`x` → `summary` → `x` would rewrite references that meant the *other* `summary`), so the refusal is deliberate.
- **A registered webhook reads as unregistered after reload** — the editor seeds from the list route, which omits `registration`. Re-registering is idempotent and returns the same URL and secret. This is also what makes the secret provably absent after reload; documented in `WebhookTriggerCard.tsx`.
- **Pre-existing, out of scope:** a project-scoped automation can't be saved on a cold load until a project is clicked (the editor reads `store.activeProjectId` instead of the record's own `projectId`). Traced to `6ffd7eca` (#465), which predates this branch — worth its own todo.
- `engine/test_support.rs` is 344 lines, over the 300 cap (already 322 at the merge base).
- Intra-step name collisions still suffix by position (`pr_list` / `pr_list_2`) — deliberate, since a step's own outputs never reorder. An `ask_agent` `expects` key colliding with a `set_variable` name isn't expressible via a per-step ordinal.
- `SetValueConfig`'s keystroke check stays scope-only — a strict subset of `validate`'s, so it never refuses a name `validate` would accept.
- M10 has no automated test (`ComposerInputPluginBridge` isn't exported); lane B skipped N4 (server-side `hookId` uniqueness) and the panel-facing last-rejection reason; lane C's M9 derives `persisted` from the store rather than a prop.

## Verification

UI typecheck clean · `TriggerTextField` 12/12, `TokenPicker` + `ConditionRow` 19/19 · prettier clean on every touched file · changeset included.


doruchiulan:
## Live QA — 4 scenarios, all PASS

Driven against the running app (Rust daemon on 31694, UI on 5267) on this branch, exercising the editor's real load → edit → save wiring rather than the pure functions CI already covers.

| # | Scenario | Result | Evidence |
|---|----------|--------|----------|
| S1 | Rename a `set_variable` output; downstream references follow | PASS | Renaming `alpha` → `beta` rewrote the later step to `value is $beta` and the hint to "Later steps use $beta."; persisted as `name:"beta"` with the notify step's `{token}` intact |
| S2 | Mid-word token round trip (braced form) | PASS | Stored `["todo/",{token}," done"]` → editor text `todo/${ticket} done` → edited to `…done now` → saved back as `["todo/",{token:{output:"value",stepId:"set-ticket"}}," done now"]`. No degradation to a literal `$name` |
| S3 | Webhook secret is register-response-only | PASS | Register surfaced a 64-char hex secret + URL + note + reveal control. The persisted registration carries only `hookId`, `url`, `lastDeliveryAt` — the `secret` key is absent, and the string "secret" appears nowhere in the stored automation |
| S4 | `ask_agent` composer-parity fields persist and reload | PASS | Selected model + permission mode, saved; persisted `{adapterId:"codex", model:"gpt-5.6-terra", permissionMode:"yolo", projectId:…}`. After a hard reload the chips re-render `GPT-5.6-Terra` / `Unattended` — the load half round-trips too |

**No surface defects.**

Two notes that are not defects in this PR:

- **The "no secret in the daemon log" half of S3 is inconclusive, not a pass.** The QA daemon log was 29 bytes, so there was nothing to search. The positive assertions above (secret absent from the stored automation, registration rebuilt field-by-field) are what actually carry S3.
- **The library list does not live-refresh for automations created outside the UI.** Rows created via `POST /api/automations` only appeared after a reload. The in-app create path updates the store locally, so this only affects API-driven setup; flagging it as an observation rather than filing it.

Also worth recording for whoever writes E2E coverage here: the model/permission menus are Radix dropdowns whose items do not respond to a programmatic `.click()` — selection has to be driven by keyboard (`ArrowDown` × N, `Enter`). My first pass read as a dropped write until I switched techniques; the fields were fine.

PR #523, opened 2026-07-25
feat(ui): smart action chips for instructions and localhost URLs (#278, #279)

doruchiulan:
Closes #278. Closes #279.

Two things agents write into chat become things you can act on.

**Instruction chips (#278).** A slash instruction in an assistant message — bare in prose, in an inline code span, or alone in a fence — renders as a chip with two actions: **Add to composer** (appends on a new line, focuses the input) and **Run in a new session** (opens an initialized draft in the source chat's project and adapter, prefilled). Neither sends. An instruction that isn't in the skills catalog stays plain text, and nothing chips while the catalog is still loading.

**Localhost tunnel chips (#279).** A localhost URL becomes a chip that opens the link when the daemon is your own machine. When it isn't, the chip tunnels the port first — a dev server on a remote daemon is one click instead of an SSH session — and warns once that the link is public. Tunnels appear in Settings → Remote Access with a stop control, and close when the chat's scope does.

## Artifacts

- Spec: `docs/specs/2026-07-25-todo-278-smart-actions.md`
- Plan: `docs/plans/2026-07-25-todo-278-smart-actions-plan.md`

## Design decisions worth a second look

**Tunnel scope is keyed by `chat_id`, path resolved at teardown** (your ruling). A tunnel belongs to the chat that started it, so a worktree moving or being removed mid-session doesn't strand a live cloudflared process.

**The daemon is the state machine; the start POST is only a trigger.** `TunnelManager::start` resolves after DNS verification (~45s), so gating the UI on it would pin `tunnelling…` long after a usable URL exists. Every transition rides a `tunnel:status` event instead. Two chips on the same port share state, but only the chip you clicked opens a window.

**Plan decision 8 was wrong about the installed `@assistant-ui/react-markdown`, and the code follows the correction.** The plan assumed a language-less fence reaches the `Code`/`Pre` overrides. It doesn't: `CodeBlockOverride` defaults `language` to `"unknown"`, so `SyntaxHighlighter` renders the body for *both* fence flavors and `DefaultCodeBlockContent` is unreachable. The plan's conclusion survives — `CodeHeader` is the one seam both flavors call, so it stays the single chip emitter — but the reasoning in the plan doc is stale on that point.

**`LinkWithPreview` was extracted out of `markdown-text.tsx`** (new file, 139 lines) to keep both files under the 300-line limit. Testids are preserved; the two `console.warn` tags moved from `[markdown-text]` to `[link-with-preview]`.

## Divergences from the plan

- **The URL chip's stop button is hidden on an errored entry**, not just while a tunnel is down. An `error` entry is a failure marker, not a live process — stopping it would POST against something already gone. The escape from an errored chip is retrying via **Tunnel and open**.
- **T17's test lives in its own file** rather than extending an existing one.
- **The `tunnelling…` badge now leads the store.** The badge derived solely from the store entry, which only exists once the daemon's first event lands — so between the click and that round trip the chip said nothing about a tunnel the user had already asked for. AC #279.1 wants the badge on click; it now reads `busy`.

## Verification

| Check | Result |
|---|---|
| `pnpm --filter @qlan-ro/mainframe-ui typecheck` | clean |
| UI suites (7 files, incl. 4 new) | 96 new tests + 46 regression, all pass |
| `cargo test --workspace` (core-rs) | 2155 pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --check` | clean |
| `prettier --check` | clean |

New suites: `instruction-actions.test.ts` (15), `instruction-chip.test.tsx` (19), `url-chip.test.tsx` (26), `port-tunnels.test.ts` (27), plus 3 FindBar regression cases — an unresolved instruction token splits a paragraph into three adjacent text nodes, and search offsets have to stay joined across them.

Regression suites re-run for every touched file: `markdown-text` (21), `TunnelControl` (3), `use-tunnel-status` (13), `reset-daemon-scoped-stores` (9), `ActivePortTunnelsSection` (5).

## Stages skipped

The reviewer/quality-reviewer panel and the QA smoke test were **not** run, to stay inside the token budget for this session. Everything above was verified by running it; nothing here rests on an agent's claim. The two most worthwhile things for a live pass: the tunnel flow against an actual remote daemon, and the instruction chip's new-session action, which is the only path that touches the draft-session machinery.


doruchiulan:
## Live QA: both instruction-chip actions fail

Manual QA pass against the running app (browser target, Rust daemon, mock replay adapter). The chips **render** correctly, but **neither action they advertise works**. Rendering was the only part of #278 that survived the pass.

### What passes

A replayed assistant message produced exactly the specified rendering:

| Case | Result |
|---|---|
| `/deploy-check` (present in the skills catalog) | chipped, `data-smart-action-token="/deploy-check"` |
| `/totally-unknown-thing` (not in catalog) | stayed plain text, no chip |
| `http://localhost:4173/` | chipped with **Open** |

### What fails

**Add to composer** — composer text never changes, focus stays on the chip. Every click throws:

```
Error: Composer is not available
    at EditComposerRuntimeImpl.setText
    at use-instruction-actions.ts  (aui.composer().setText)
    at onClick (InstructionChip.tsx)
```

**Run in a new session** — the toast surfaces a raw internal error:

```
Couldn't start a new session
useClientLookup: Index 1 out of bounds (length: 0)
```

### Root cause (confirmed, not inferred)

`useInstructionActions` calls `useAui()`, but the chip renders **inside a message**, so it receives a *message-scoped* proxy rather than the thread's. Deferred `aui.composer()` calls resolve through that message's index in the *currently active* thread:

- `append` → resolves to the message's `EditComposerRuntimeImpl`, which only exists while that message is being edited → "Composer is not available". The `composer.__internal_getRuntime?.()` fallback goes through the same scoped proxy, so it can't rescue it.
- `runInNewSession` → `switchToNewThread()` makes an empty thread active; the scoped lookup then asks for message index N in a list of length 0.

Verified by prediction: with the chip in the assistant message at index 1 the error reads `Index 1 out of bounds`; re-seeded so the chip lands at index 3, the same click reports `Index 3 out of bounds`. The index tracks the chip's message position.

This is why the unit suites are green — they mount the hook outside a message context, where the proxy resolves to the thread composer. The failure exists only in the real render tree.

### Suggested direction

Resolve the thread composer explicitly rather than through the message-scoped `aui` (and capture it before `switchToNewThread()` in the new-session path, since the scope is invalidated by the switch).

Separately: `runInNewSession` puts a raw internal error string into a user-facing toast. Worth a friendlier message regardless of the fix.

### Not covered

The URL chip's **Open** action (#279) and the FindBar split-token regression were not exercised in this pass. The remote-daemon tunnel flow was not tested — it needs a paired remote daemon, out of scope here.

*Manual QA, no source changes made on the branch.*

doruchiulan:
Two findings from live testing, one fixed in `1f0aa80d`.

**1. Both chip actions were writing to the wrong composer — fixed.**

`useInstructionActions` called `aui.composer()`. Chips render inside a message, and `MessageByIndexProvider` rebinds the aui context's `composer` accessor to `aui.thread().message({index}).composer()` — that message's *edit* composer, not the thread's.

- **Add to composer** silently wrote to a `NoOpComposerClient` (the message is not being edited), so nothing appeared.
- **Run in a new session** additionally threw: after `switchToNewThread()` the new main thread has zero messages, so the lazily-evaluated `message({index: N})` lookup blew up at `useClientLookup.js:44` and landed in the `catch` as `Couldn't start a new session — useClientLookup: Index N out of bounds (length: 0)`. The index varied with the chip's message position, which is what made it look random.

The fix routes both actions through `runtime.thread.composer`. `threads.main` is bound to a `{type: "main"}` selector and `ThreadComposerRuntimeImpl` recomputes its state whenever the subject is dirty or unconnected, so it resolves the live composer on both sides of the switch — and the `__internal_getRuntime` stale-read workaround is no longer needed.

Why CI stayed green: `__tests__/instruction-actions.test.ts` mocked `useAui()` with a thread-composer-shaped stub, so the message scoping never existed in the test. The mock now models it — `useAui().composer()` throws the real `useClientLookup` error, and `useAssistantRuntime()` carries the thread composer. Reverting the implementation fails 6 of the 15 tests.

Verified live (browser mode): Add to composer inserts and appends on a new line; Run in a new session opens a draft with the instruction prefilled and no toast.

**2. Right-click on a chip falls through to the host's native menu.**

Chips are `<span>`/`<button>` with no `ContextMenu` wrapper, so a right-click is not `preventDefault`ed (confirmed live on `smart-action-instruction-append`) and the browser's own menu appears.

For instruction chips this is not a regression — that text was plain before. For **localhost URL chips it is**: those links previously rendered as `LinkWithPreview`, which carries *Copy link* / *Open link* on right-click. Spec decision 14 deliberately dropped the Copy button from the chip ("the URL stays selectable text"), so this is the cost of that decision rather than a coding slip — flagging it because the lost affordance is a right-click away, not a visible one. Say the word if you want the chip wrapped in a context menu with Copy URL.

PR #525, opened 2026-07-25
feat(ui): in-thread context menus for file paths and text selections (#274, #280)

doruchiulan:
## Summary

Two in-thread right-click menus, built on one shared substrate:

- **#274** — right-click a file path in a tool-result header → **Copy Absolute Path** / **Copy Relative Path**.
- **#280** — right-click a text selection in an assistant message → **Start new session** / **Add to composer**. Neither auto-sends.

Anywhere neither trigger applies (prose, an image, empty space), the native WKWebView menu is left alone.

- Spec: `docs/specs/2026-07-25-todo-274-thread-context-menus.md`
- Plan: `docs/plans/2026-07-25-todo-274-thread-context-menus-plan.md`

## Decisions the user made at the design gate

These were ruled on interactively during prototyping; they are recorded here because there is no later gate where they surface.

- **Merged menu, not either/or.** When a selection is active *and* the right-click lands on a file path, one menu shows both sections rather than the path actions winning. The user: *"it's C for sure, we already have the same behaviour for quote."*
- **Quotes stack across messages.** "Add to composer" appends a segment; quoting from a second message keeps the first. This is the ruling that forced the composer's segment model — the composer is no longer a single string but an ordered list of quote and text segments, serialized on submit.
- **The composer text stays editable** with pills in it, and prose inside a selection menu's source message stays selectable.

## Planner disagreements

The planner objected to the segment-model rebuild as disproportionate to a "quote" feature, preferring a plain string append with a `> ` prefix. Overruled by the cross-message stacking decision above: a string append cannot represent two quotes with distinct provenance, and the user asked for the stacking behavior explicitly.

## Review verdicts

`reviewer` and `quality-reviewer` both ran on `git diff main...HEAD`; must-fix findings are addressed in `a6173031`. Notable fixes:

- **Copy feedback was lying.** The menu said "Copied" unconditionally. `writeToClipboard` now returns a boolean, and a rejected write reads **Copy failed** with the clipboard untouched.
- **Abandoned drafts leaked quotes.** assistant-ui reuses one `__LOCALID_*` slot until first send, so quotes pinned into a draft survived into the next new thread. `resetNewThreadDraft` now clears the segment store for that slot.
- **The empty menu ate the native one.** With nothing to copy, a disabled menu still rendered and suppressed WKWebView's menu. The Radix trigger is now suppressed outright so the native menu survives.

### Deliberately not fixed (nice-to-haves)

- Autoscroll during streaming dismisses the selection toolbar. Pre-existing behavior of the scroll container, not introduced here.
- `ref.absolute ?? ref.relative` when `bases` is `{}` — unreachable through the UI; every call site supplies a base.
- `file-ref.ts` does not normalize `..` segments. No producer emits them today.
- `markdown-text.test.tsx` (376 lines) and `Composer.test.tsx` (362) exceed the 300-line limit. **Both were already over at the merge base `4ce69b3a`**; splitting them is unrelated churn.

## Tests

Three previously-skipped selection E2E specs are **un-skipped**. The skips blamed the app; the root cause was a test bug — the TreeWalker's first "project" text node was inside the reasoning block, outside `.aui-md`, where `body { user-select: none }` applies. `addRange` reported `rangeCount === 1` while `getSelection().toString()` stayed empty.

`de99039c` fixes a genuine blocker this branch introduced: `tool-cards.spec.ts:235` (added by Phase B) fails under `E2E_MODE=mock`, which is what `.github/workflows/e2e-mock.yml` runs. The mock recordings bake the recording machine's absolute project path, and `remap_history_paths` only rewrites rehydrated history — never the live sink stream, which is what this spec reads. Fixed with a `{{PROJECT_PATH}}` placeholder resolved against `SessionOptions.project_path` at fixture load, so both assertions stay honest (absolute copy asserts `${project.projectPath}/index.ts`, relative asserts `index.ts`).

## QA

12 scenarios, **12 PASS, 0 defects**, run against an isolated daemon:

Both copy items present on a path (`itemCount:2`) · absolute copy · relative copy · native-menu fall-through on prose · fall-through on a pill with an active selection · quote-without-send (user-message count unchanged) · **cross-message quote stacking** (2 pills, distinct `data-segment-id`s, two source messages) · ordered serialization on submit · "New session" prefills a draft and creates no chat (daemon chat count 1 → 1) · draft-leak guard (0 → 1 → abandon → 0 → New → 0) · truthful copy failure (`writeText` rejecting `NotAllowedError` → "Copy failed", clipboard still holds its prior value).

Three honest deviations:

1. The project-switch variant of the leak test was not run separately — only one project existed in the isolated data dir.
2. Native-menu fall-through was proven by proxy (`event.defaultPrevented === false` and zero `[role=menu]` elements), not by visually confirming the macOS menu, which is out of the webview's reach.
3. The QA environment was hand-built. The agent launch scripts still boot the retired Node daemon; that fix is #524, on its own branch, and cherry-picking it here would have polluted this PR.

## Note for anyone running this branch locally

`packages/ui/dist` was rebuilt with `VITE_DAEMON_PORT=31416` baked in for the QA harness. Rebuild before `tauri:dev` or packaging.

Closes #274
Closes #280


doruchiulan:
## Live QA — 4 scenarios, 4 PASS, no defects in this PR

Ran against the running app (Rust daemon + renderer, browser target) on this branch merged with latest `main`, using a deterministic `mock-cli` NDJSON fixture and a throwaway project. This is manual-QA of the running surfaces only — CI owns the unit/typecheck/lint side.

| # | Scenario | Result |
|---|----------|--------|
| S1 | Quote a selection with prose already typed | **PASS** — main composer cleared, quote captured, and the typed `PROSE_ONE` preserved as a committed segment's `<textarea>` value (not lost) |
| S2 | Two quotes interleaved with prose → send | **PASS** — serialized in exact order, both quotes emitted as real markdown blockquotes |
| S3 | "New session" from a selection | **PASS** — draft opened on the source project/branch, composer prefilled with exactly the selection, segments cleared |
| S4 | Right-click a tool-card path | **PASS** — Copy Relative → `index.ts`, Copy Absolute → the full path; and with a text selection active the Radix menu correctly does **not** open, so the native menu survives |

S2 detail — the sent message serialized as:

```
PROSE_ONE

> ALPHA_SELECTION_ONE is the first quotable sentence.

PROSE_TWO

> BRAVO_SELECTION_TWO is the second quotable sentence.

PROSE_THREE
```

Verified as two `<blockquote>` elements in the correct positions, so the `> ` prefixing and ordering both hold end-to-end, not just in the pure serializer's unit tests.

### Why S3 got extra scrutiny

#523 ships a structurally similar "Run in new session" action from inside an assistant message, and **it fails 100% of the time** there — its hook calls `useAui()` from a component rendered *inside a message*, so the proxy resolves to that message's `EditComposerRuntimeImpl` instead of the thread composer (details in #523).

This PR does not have that bug, for two concrete reasons:

1. `ChatSelectionToolbar` is mounted at **thread level** (`ChatThread.tsx:128`), not per-message, so its `useAui()` resolves to the thread's composer.
2. The order-sensitive sequence is extracted into the shared, dependency-injected `openNewThreadDraft` — which #523 hand-rolled a second copy of instead of reusing.

Worth flagging across the two PRs: whichever merges second should probably move #523's chip actions onto this `openNewThreadDraft` helper rather than keeping the duplicate.

### One pre-existing issue noticed (NOT from this PR)

Rendered user-message paragraphs carry a stray `node="[object Object]"` attribute in the DOM. Source is `UserMessage.tsx:85` — `DirectiveParagraph` does `<p {...props}>`, spreading react-markdown's `node` prop into the element. That file isn't touched here; this PR only makes it easier to see, because user messages now routinely contain multi-block markdown. Invalid-attribute noise, no functional impact — filing as an observation, not a request to fix in this PR.

### Not covered

Editor context menu (`EditorContextMenu.tsx`), nested-transcript path menus, and the attachment-adapter changes were not exercised in this pass.

PR #535, opened 2026-07-28
feat: skills management UI in Setup Advisor (#243)

doruchiulan:
## Summary
Adds a Skills section to the Setup Advisor dialog: a section switcher beside the title (Recommendations / Skills), listing the active project's skills grouped by scope, with search, inspect (raw SKILL.md content), and delete (project/global scope, behind a confirmation). Deleting revalidates the composer's `/` picker and the sidebar Skills tab, which now links into the section.

Artifact: `docs/plans/2026-07-28-todo-243-skills-management-ui-plan.md`

Review: APPROVED (1 round)
QA: 11/11

## Decisions
- **D1**: `npx skills` CLI detection is implemented as `skills-lock.json` presence detection (via the existing files route), not a real process probe — no Tauri exec command exists and the todo's AC forbids new daemon endpoints. A true installed-CLI check needs a Tauri exec command or daemon capability route.
- **D2**: New testids use `setup-advisor-*` / `skills-section-*`; legacy `automation-recommender-*` prefix untouched.
- **D3**: Recommendation report still fetches on every open, even when opening straight onto Skills — keeps `SetupAdvisorHost`'s existing effect and its test file byte-identical.
- **D4**: Scope handling ships as grouping (`SectionHeader` blocks), not a separate filter control.
- **D5**: Inspect view renders raw content in a `<pre>` — no markdown renderer, so no `React.lazy` needed.
- **D6**: "Adapter has no skills support" detected by matching the daemon's exact 404 body string, pinned as an exported `UNSUPPORTED_ERROR` constant.
- **D7**: `resetDaemonScopedStores` bumps the new revalidation nonce rather than zeroing it, so a daemon switch always forces a refetch.
- **D8**: Delete affordance gated by scope + backing file only, never by adapter id — the delete route is Claude-only while list also serves mock-cli, and the brief puts that gating difference out of scope.
- **D9**: `docs/plans/` is gitignored in this repo; the plan was force-added with `git add -f`, matching treatment of other already-tracked gitignored plan docs. Flagging for the repo owner to consider un-ignoring the path.
- **D10**: Found and fixed a latent bug the section work would have activated — `MainToolbar.tsx`'s `onClick={openSetupAdvisor}` would have passed the click event as `openSheet`'s new section argument. Fixed with an explicit call plus argument normalization, pinned by a test.
- Additional implementation-time decisions (act() wrapping for a red test file, SkillsSection decomposition to stay under 50 lines/function, testid reuse between row and inspect header, empty-vs-no-results copy split, CHIP_BASE extraction with two design-system doc pointer updates) are recorded in the plan/implementation log.

## Test plan
- [x] `pnpm --filter @qlan-ro/mainframe-ui typecheck` clean
- [x] All new/updated unit tests green (11/11 QA)
- [x] No sibling-group file collisions during implementation (verified via `git diff --cached --stat` before each commit)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


doruchiulan:
Rejected — the scope was wrong, not the code.

This built a skill viewer and deleter. What #243 needs is what skills.sh is: install and uninstall skills, and nothing else.

The plan's D1 decision resolved `npx skills` by telling the user which command to run. A button has to run it. Doing that needs an execution path the previous brief forbade, so #243 goes back for a re-brief that explicitly authorizes one — a Tauri shell command or a daemon endpoint.

Branch kept on origin so this diff stays readable.

PR #538, opened 2026-07-29
feat(claude): offer the older models the CLI's picker hides

doruchiulan:
## The problem

The Claude CLI's `initialize` control response returns only the five models its own picker shows — Default, Opus (1M), Fable, Sonnet, Haiku. A successful probe **replaced** the static catalog outright, so every older model the API still serves was unreachable from Mainframe: Opus 4.8/4.7/4.6/4.5/4.1 and Sonnet 4.6/4.5.

## The fix

Merge instead of replace. `merge_older_models` appends a curated `OLDER_MODELS` list to the enriched probe result, skipping anything the probe already offers — matched by id **and** by `resolvedModel`, with a `[1m]` suffix normalized away so the `sonnet` alias doesn't double up with `claude-sonnet-5`.

The list was checked against [the model-deprecations page](https://platform.claude.com/docs/en/about-claude/model-deprecations) on 2026-07-29. `claude-opus-4-20250514` and `claude-sonnet-4-20250514` are deliberately **excluded** — both retired June 15, 2026, so offering them would 404. `the_catalog_lists_no_retired_model` asserts that, and is the test to update when the list next moves.

Same pass corrects the static fallback catalog, which had drifted: the "Default - Opus 4.8" entry is now Opus 5, and `claude-opus-5`, `claude-fable-5`, 4.8 and 4.7 were missing entirely.

## Shape of the change

- **New `models.rs`** in `mainframe-adapter-claude` — the whole catalog surface (static list, `enrich_with_context_window`, the merge) moved out of `adapter.rs`, which was over the 300-line limit.
- **New `isOlder` flag** on `AdapterModel`, in both the TS and Rust twins.
- **Composer picker** splits current from older rows, with an "Older models" label and a `max-h-[264px]` scroll cap. Twelve undifferentiated rows in a `w-72` popover would have been worse than the problem.
- **Automations `ModelMenu`** got a scroll cap too — it's a flat all-providers list, so it grows with every new entry.
- Settings' default-model select picks the new entries up for free (`buildModelOptions` spreads `adapter.models`).

Not touched: `packages/core/src/plugins/builtin/claude/*`, the orphaned TS daemon that CLAUDE.md keeps only for its version field.

## Verification

- 337 `mainframe-adapter-claude` lib tests pass — 23 in the new `models` module, 6 of them new (merge dedup by id, dedup by resolved alias, probed entries left untouched, retired-id guard, older-flag correctness).
- `cargo check --workspace --all-targets` and `cargo fmt --check` clean.
- 20 `ProviderModelSelect` tests pass, 4 new: the label renders only when an `isOlder` model exists, older rows sort after current ones, selecting one calls `setModel` with its exact id.
- `pnpm --filter @qlan-ro/mainframe-ui typecheck` clean.


doruchiulan:
## Follow-up: model metadata now comes from the CLI registry (55c5b4e)

`docs/adapters/claude/MODELS.md` prompted a re-check of the first commit's metadata. That doc is v2.1.112-era (reconciled 2026-04-17) and stale in places, so I verified against the installed CLI binary's own model registry instead — it carries `context: {window, native_1m, …}` and `capabilities: [effort, max_effort, xhigh_effort, adaptive_thinking, fast_mode, …]` per model.

The first commit inferred window and effort ladder from the model *family*, and got five wrong:

| model | was | CLI v2.1.220 |
|---|---|---|
| `claude-opus-4-8` | 200k window | `window: 1e6, native_1m: true` |
| `claude-opus-4-7` | 200k window | `window: 1e6, native_1m: true` |
| `claude-sonnet-5` | no xhigh | has `xhigh_effort` (so: ultracode) |
| `claude-opus-4-6` | xhigh | no `xhigh_effort` |
| `claude-opus-4-5` / `sonnet-4-5` / `opus-4-1` | full ladder | 4.5 stops at high; the other two have no effort support |

The 1M ones are the ones that mattered — `enrich_with_context_window` seeds probed entries from this table, so the context indicator would have under-reported Opus 4.8/4.7 by 5×.

The effort constants are renamed `EFFORTS_TO_XHIGH` / `_TO_MAX` / `_TO_HIGH` — for the rung they reach rather than the family they were assumed to fit, since the family is exactly what misled here. `model_windows_and_effort_ladders_match_the_cli_registry` pins the table.

Fast-mode and adaptive-thinking flags were already correct against the registry; no change there.

### Unrelated: one line that unbreaks main

`cargo check --workspace --all-targets` is red on `main` — `permission_cancel_tests.rs` declares `tracker_end_all_running` twice in a single impl block (came in with #529). Deleted the duplicate here so this PR can be verified; it's one line and touches nothing else.

PR #600, opened 2026-08-10
test(e2e): unblock the release gate — pin the transcript pair, fix the chat-selection race, raise the mock budget

doruchiulan:
Closes the remaining rc.22 e2e failures, one by fixing it and one by pinning it to a filed bug. Supersedes the earlier "wait for the transcript" approach in this branch, whose premise the Linux evidence disproved.

## 1. The transcript pair is an app bug, not a test bug — todo #320

`e2e-mock` has `workflow_dispatch`, so I instrumented the two tests and ran the suite on Linux. The transcript **empties partway through the file and does not come back**: the four tests above pass while asserting message content, then every later read of the thread returns a mounted, empty viewport for the full 15s the probe polls:

```
{threads:1, messages:0, textParts:0, scrollHeight:636, clientHeight:636}
```

So `chat-scroll-to-bottom` had nothing to enable for, and Find's `0/0` was the honest answer. **⌘F is not broken** — worth saying, because for two releases the artifacts made it look like it might be.

The clue for whoever picks up #320: in the failing snapshot the sidebar holds exactly one session, titled **"Untitled session"**, active and empty — a session with no messages, not one whose view was lost. The daemon doesn't persist message history (the CLI replays it via `--resume`), so a client-side re-seed would produce exactly this. Plausible, not confirmed.

Both tests are skipped against #320 with the evidence in the file. They keep their preconditions, so the run after the fix reports what was missing rather than a disabled button. A second defect is recorded on the way past: on the one attempt that saw a populated transcript, the scroll button went enabled → invisible between the assertion and the click, so a programmatic `scrollTop = 0` doesn't durably disengage stick-to-bottom.

## 2. `createTauriChat` now confirms the chat it hands back is active

`tool-cards` plan-approval failed both attempts on Linux while passing locally. Its screenshot shows the sidebar row for the chat under test reading **"waiting"** — the plan gate genuinely existed — while the visible thread was a different chat. The gate was rendering out of view.

The helper clicked the new chat's row and returned without checking the selection held; the click and the plan-mode toggle each fire `chat.updated`, and the reload that follows can revert the active thread. Nothing errors when that happens — the caller's next `sendMessage` just goes to the other chat. The helper now verifies and re-takes the selection. A no-op when it never moved, and it hardens every spec that creates a chat.

## 3. Mock budget 45s → 60s

At 45s the budget exactly tied `tool-cards`' own `waitFor({ timeout: 45_000 })`, so the backstop killed the test first and reported "Target page has been closed" instead of "waiting for chat-plan-gate". The failure was real either way, but a backstop below an in-test wait buys nothing and costs the error message — the opposite of what #598 was for. My regression, fixed here.

## Verification

- Full local suite green on the transcript/budget changes (385 passed).
- Full local suite with the helper change: one unrelated `ENOTEMPTY` in another spec's `afterAll` teardown; that spec passes 2/2 in isolation and the change touches neither worktrees nor cleanup.
- Linux `workflow_dispatch` run in flight; result to follow on this PR.

### DIFF
--- a/.changeset/plenty-lizards-lick.md
+++ b/.changeset/plenty-lizards-lick.md
@@ -1,0 +1,4 @@
+---
+---
+
+E2E-only: wait for the transcript to render before the scroll and find tests measure it. No shipped behavior changes.
--- a/packages/e2e/helpers/tauri/setup.ts
+++ b/packages/e2e/helpers/tauri/setup.ts
@@ -2,7 +2,7 @@ import { execSync } from 'child_process';
 import { mkdtempSync, mkdirSync, writeFileSync, realpathSync, rmSync } from 'fs';
 import { homedir } from 'os';
 import path from 'path';
-import type { Page } from '@playwright/test';
+import { expect, type Page } from '@playwright/test';
 import { DAEMON_PORT } from '../../fixtures/daemon.js';
 import { sessionsSidebar, composer } from './page-objects.js';
 import { waitConnected } from './wait.js';
@@ -86,6 +86,19 @@ export async function createTauriChat(
     await toggle.waitFor({ timeout: 10_000 });
     if ((await toggle.getAttribute('aria-pressed')) !== 'true') await toggle.click();
   }
+
+  // Confirm the new chat actually HOLDS the selection before handing it back.
+  // The row click and the plan toggle each fire `chat.updated` → a thread-list
+  // reload that can revert the active thread to whatever was open before (the
+  // race documented on the plan-approval tests). Nothing failed when that
+  // happened — the caller's next `sendMessage` simply went to the other chat,
+  // and the assertion that came after looked for a gate rendering out of view.
+  // Re-taking the selection is a no-op when it never moved.
+  await expect(async () => {
+    if ((await row.getAttribute('data-active')) !== 'true') await row.click();
+    await expect(row).toHaveAttribute('data-active', 'true', { timeout: 2_000 });
+  }).toPass({ timeout: 20_000, intervals: [250, 500, 1_000] });
+
   return chatId;
 }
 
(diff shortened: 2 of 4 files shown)


doruchiulan:
Linux `workflow_dispatch` run: **green** — [run 31408665782](https://github.com/qlan-ro/mainframe/actions/runs/31408665782). 384 passed, 0 failed, 2 flaky (`find-in-path`, `sessions-draft` history-read — both recovered on retry), 39 skipped. Job 11m46s, test phase 6.6m.

Caveat worth recording: `tool-cards` passing here is consistent with the `createTauriChat` fix but does not isolate it — that test also passed on the Linux run *before* the fix, and failed both attempts on the one after. One green run is evidence, not proof, for that particular change.

doruchiulan:
Reverted the `createTauriChat` selection guard (ee5d3239).

Reasoning: it never demonstrably fixed `tool-cards:402` — that test failed **with** the guard in place in two later runs — and `find-in-path` has now failed in 3 of the 4 Linux runs since the guard landed, having never failed in any run before it. The guard adds a click and a poll to setup that every spec shares, which fits failures clustering on the first tests of a describe. Unproven fix plus suspected new breakage is not a trade worth keeping.

That leaves this PR as: transcript pair skipped against todo #320, and the mock budget 45s → 60s. `tool-cards:402` reverts to what it was before this session — a load-sensitive failure that also failed in the run *preceding* any of my changes.

doruchiulan:
Superseded by #607, which fixes both transcript tests properly instead of skipping them.

What changed since this PR was written: the instrumented run showed the `find-in-chat` `0/0` was a *downstream* effect — the scroll test failing triggers a retry, the retry re-runs `beforeAll` into a fresh empty chat, and every later test in the describe then runs against it. So the skip this PR proposed was treating a symptom of a symptom. #607 fixes the scroll test's real failure (the viewport re-anchoring between the enabled assertion and the click) and declares the describes serial so a retry re-runs the sends.

The one line worth keeping — the 45s → 60s mock budget — has been carried into #607. The `createTauriChat` guard was already reverted here after it turned out to break `find-in-path` in 3 of 4 runs while never fixing what it targeted.

This branch is also ~100 files behind main after the assistant-ui 0.15 migration (#602), so rebasing it to keep a now-unnecessary skip isn't worth it.

PR #603, opened 2026-08-10
fix(ui): refuse an empty background history re-seed (does not close todo #320)

doruchiulan:
The product bug behind the two `transcript.spec.ts` failures that have been red since rc.20. **This is a real user-facing fix, not a test fix** — and it's the reason those tests were failing, which for two releases read as "⌘F is broken" and "the scroll button doesn't work".

## What happens

`history.loaded` replaces the transcript wholesale, and a re-seed fires on every reconnect and every reattach after dormancy (`chat-ws-subscription.ts` → `refreshInBackground`). One empty payload at the wrong moment blanks a populated thread, and nothing schedules another read to undo it — so it stays blank until the next live message arrives.

## Why the fix is here and not in the daemon

I instrumented `ChatManager::get_messages` and ran the suite on Linux. Across the whole run:

- `load-failed` — **0 times**
- `disk-empty` — **0 times**
- `no-history-session` — **every** empty response

So the daemon is not failing a read and reporting it as empty; it genuinely has nothing for a chat whose CLI session it can't build a read from yet. "Empty" there means *"I have nothing for you"*, and only the client knows it already has a transcript on screen. My earlier plan — have the daemon distinguish a failed read — would never have fired.

## The guard

A **background** re-seed returning zero messages for a thread that holds messages is refused, and settles `loadState` so nothing strands on a spinner.

Two things it deliberately does not break:
- The **first** load is never refused, so a genuinely empty thread still renders as one.
- A real emptying still lands: `/clear` arrives as its own `messages.cleared` event, not as a silent empty read.

## Verification

Three unit tests, each verified to **fail without the guard** (I removed it and watched them go red — a regression test that passes either way is worthless, which I learned the hard way on a different change today). Full controller suite: 252 passing.

The two `transcript.spec.ts` tests are the end-to-end acceptance test. They are unskipped on `main`, so the Linux run on this branch exercises them directly — if they pass, this fix is confirmed against the failure it was derived from, and the skip proposed in #600 can be dropped.

Related: todo #320 carries the full investigation, including two hypotheses this disproved.

### DIFF
--- a/.changeset/cold-toes-rush.md
+++ b/.changeset/cold-toes-rush.md
@@ -1,0 +1,5 @@
+---
+'@qlan-ro/mainframe-ui': patch
+---
+
+Stop a background history re-seed blanking a transcript that is already on screen. A reconnect or reattach re-reads history and replaces the thread wholesale, and the daemon answers "empty" for a chat it has no CLI session to read from yet — so one badly-timed re-seed emptied a populated thread until the next message arrived.
--- a/packages/ui/src/features/chat/controller/__tests__/chat-thread-controller-load.test.ts
+++ b/packages/ui/src/features/chat/controller/__tests__/chat-thread-controller-load.test.ts
@@ -17,6 +17,7 @@
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import type { DaemonWsClient } from '../../../../lib/daemon/ws-client';
+import type { DisplayMessage } from '@qlan-ro/mainframe-types';
 import type { DaemonEvent } from '@qlan-ro/mainframe-types';
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


doruchiulan:
**Correction — this fix does NOT close the transcript tests, and the PR description above overclaims.**

The two `transcript.spec.ts` tests are unskipped on this branch, so [run 31438658172](https://github.com/qlan-ro/mainframe/actions/runs/31438658172) was their acceptance test. They still fail, identically: `find-in-chat` still reports `0/0`.

So the destructive re-seed is real — the unit tests prove the guard refuses an empty background payload, and they fail without it — but it is **not** the path that empties the transcript in that spec. Something else does.

What that means for the two claims in the description:
- "This is what the two skipped tests were catching" — **wrong**, retracted.
- "if they pass, the fix is confirmed" — they did not pass, so this fix is confirmed only at the unit level: a correct guard against a real path, of unproven user impact.

I would still merge it on its own merits (it closes a genuine way to lose a transcript, cheaply and with tests), but not as a fix for todo #320. #320 stays open and the skip in #600 stands.

doruchiulan:
Rebased onto current main (post assistant-ui 0.15). Controller suite re-run against 0.15: **252 passed**, typecheck clean.

Recommendation stands: merge as defensive hardening, not as a fix for todo #320. The path it guards is real — an empty background re-seed *would* wholesale-replace a populated transcript, and the three unit tests fail without the guard — but the failure that motivated it turned out to be something else entirely (see #607). Small, tested, and it can only prevent a class of data loss.

PR #655, opened 2026-08-15
docs: rewrite API reference for the Rust daemon

doruchiulan:
## Summary

`docs/API-REFERENCE.md` documented Express routes from `packages/core/src/server/routes/`, a package that has zero files left in it — the HTTP/WS server was fully ported to Rust (`packages/core-rs/crates/mainframe-server`, Axum). This replaces the doc entirely, built directly from the current route files rather than carried over from the old one.

- **Route coverage**: all 130 distinct `/api/*` paths across the 41 files in `mainframe-server/src/routes/`, verified by diffing every `GET|POST|PUT|PATCH|DELETE` path string documented against every path string in the source (`grep -o '"/api...' *.rs`) — zero missing, zero extra. Counting individual method registrations (a route with `.get().post()` counts twice) gives 150 HTTP endpoints from the route files, plus `GET /health` and the plugin manager's two generic routes (`GET /api/plugins`, `GET /api/plugins/{id}`) — 153 first-party endpoints total.
- **Response envelope**: verified against `respond.rs` (`ok`/`ok_empty`/`fail`) and the canonical `ApiResponse<T>` type in `packages/types/src/api.ts` — they match exactly. Documented every deviation from the standard envelope I found while reading the route bodies: `/health`'s bare JSON, `POST /api/projects`'s 409-with-`data`, `DELETE /api/tags/{name}`'s bodyless 204, the webhook ingress status mapping, quota's `okEmpty`-means-"no data yet" convention, automations' 202-on-start, and git-write's raw-error-message 500s.
- **WebSocket protocol**: documented from `websocket.rs` + `ws_schemas.rs` + the canonical `ClientEvent`/`DaemonEvent` enums in `packages/types/src/events.ts` — all 6 client→server event types and all ~50 server→client event types, the subscribe handshake order, and the connection-global fan-out gating.
- **Auth**: documented the loopback-bypass model, the `AUTH_TOKEN_SECRET` no-op-when-unset behavior, the auth-exempt paths, and the WS upgrade's separate `?token=` query-param auth (never the header).

## What I couldn't verify or flagged as out of scope

- Per-plugin sub-routes nested under `/api/plugins/{id}/*` are plugin-defined and dynamic — not enumerated, pointed at the plugin's own router instead.
- `launch.json`'s full configuration schema and the Automations v2 step-graph body shape are large enough to warrant their own reference; this doc gives the route surface and points at the parser/type source instead of reproducing the full schema.
- `GET /api/settings/general` and `/providers` response bodies are assembled as a `serde_json::Map` field-by-field, not a single named struct, so they're described in prose rather than cited against a type.

## Surprises worth flagging

- **No HTTP surface exists for the Claude CLI `/workflows` feature.** `chat_workflow_runs.rs` looked like a route file from its name but is an internal disk-backfill helper `chats.rs` calls — never mounted as a router. The `workflow.*` WebSocket events are real (Automations v2), but easy to mistake for that older feature; called this out explicitly since the old doc's groupings wouldn't have caught it.
- The 409-with-`data` on `POST /api/projects` and the bodyless 204 on `DELETE /api/tags/{name}` are both genuine, deliberate deviations from the envelope — not bugs, but worth a client author knowing about up front.
- The daemon binds `127.0.0.1` only, always — remote/mobile access goes exclusively through the cloudflared tunnel, never a wider bind.

## Test plan

- [x] Every documented path cross-checked against `grep`-extracted route paths from the 41 route files (script run, zero diff in either direction)
- [x] Every JSON code fence parses (`json.loads` over each fenced block)
- [x] Docs-only change; `pnpm changeset --empty` (hand-written — fresh worktree had no `node_modules` to run the CLI)


doruchiulan:
Follow-up commit fixes a handful of request-shape claims (query-param names, body field names) I'd written without reading the actual struct — caught in a second accuracy pass. Path coverage was already verified mechanically and didn't change; the fixes are: `files.rs` query-param names + the `PUT /api/projects/{id}/files` body-vs-query shape, `git_chat.rs`'s `files` field name (not `paths`) on stage/unstage/commit, `GET /api/chats` and `GET /api/automations` filter query params, and the automation-credentials PUT body field (`token`, not `value`).

PR #669, opened 2026-08-16
fix(ui): unwedge the session hover card, and make the plan-revision e2e retry-safe

doruchiulan:
Fixes the three failures that held the v2.0.0 e2e gate red (run [31957985134](https://github.com/qlan-ro/mainframe/actions/runs/31957985134/job/95191484881): 2 failed, 1 flaky).

## 1 + 3 — `sessions-rows.spec.ts` — one product bug, two symptoms

Radix opens a hover card on **focus** as well as `pointerenter`, from a 500ms timer. Clicking a row focuses its button, so the meta card popped open half a second later with the pointer long gone, and only a stray pointer move (the existing `useHoverCardWedgeGuard`) ever closed it again.

Reproduced live, before any fix (throwaway probe against the mock daemon):

- after the setup clicks + `mouse.move(0, 0)`: **1 card open**, neither row `:hover`, `document.activeElement` a button inside a row — a click-focused wedge;
- hovering one row then produced **2 open cards**, 48px apart, i.e. one per row → the strict-mode violation in *"hover card names the row's project"*;
- right-click, wait past 500ms: the card opened **over** the context menu. Radix's `DismissableLayer` only lets its **top** layer answer Escape (`react-dismissable-layer/dist/index.mjs:90-107`), so Escape closed the card and left the menu up — Escape #1 menu still open, #2 still open. That is *"right-click context menu … before any message has been sent"* failing `toHaveCount(0)`. CI's own error-context snapshot shows both the open `menu [active]` and the hover card in the same tree.

The row now refuses an open the pointer never asked for — `:hover`, or a keyboard `:focus-visible` inside the row, so click-focus is out and keyboard users keep the card — and suppresses the card outright while its context menu is open. After the fix the same probe reports 0 wedged cards, 1 card on hover, 0 cards while the menu is up, and Escape #1 closes the menu.

Both tests are hardened to reproduce the CI timing locally instead of racing it: `:193` now waits past the 500ms delay and asserts no card opened, `:241` names the duplicate (`toHaveCount(1)`) rather than reporting a locator error.

## 2 — `tool-cards.spec.ts` plan revision — two mock-replay bugs

**The retry could never pass.** `MockCliAdapter` hands out fixtures from a per-daemon counter (`<key>.<N-1>.ndjson` for the Nth session), so a test needing the *second* recording was only correct while an earlier test in the same describe had consumed the first. A Playwright retry re-runs the failed test alone: the counter restarts at 0 and the retry replayed the **approved-plan** recording, which never opens a second plan gate. CI's retry snapshot shows exactly that — the approved-path transcript plus an `Untitled session` row for the chat that never got a message. So a single first-attempt flake here was unrecoverable and failed the whole run. The revision tests (this one and its twin in `chat.spec.ts`) now get their own describe, daemon, and recording key — `plan-approval.1.ndjson` → `plan-revision.0.ndjson` — which also retires the mid-test `createTauriChat` navigation race they both carried.

**The first attempt** died on `mock-cli: expected a 'respondToPermission' marker but the fixture had nothing (fixture exhausted)` (visible in the CI failure screenshot): a second permission response reached the replay after the recording's single deny marker was consumed, the run errored, and the second gate never mounted. A `respondToPermission` past the end of a recording is now ignored rather than fatal — the real CLI drops a control response for a request it already resolved, and recordings routinely stop at the last gate they opened, so erroring made the double stricter than the thing it doubles. Same tolerance `take_interaction` already gives an in-recording duplicate; every other desync (a stray `sendMessage`) stays fatal.

**Not fixed here, worth a follow-up:** the origin of that extra deny-path response is unproven — 959 local daemon logs show zero duplicate `permission.respond` frames from the client, and it did not reproduce in ~25 local runs including 8× CPU throttling. What *is* proven is a daemon-side double-forward on the **allow** path: `permission_handler.rs:288` forwards the response, then the plan-mode escalation forwards the same one again (`plan_mode_handler.rs:34`) — which is why `plan-approval.0.ndjson` carries two identical `respondToPermission` markers 17ms apart. Benign against the real CLI, but it deserves its own issue.

## Verification

- `sessions-rows` + `tool-cards` + `chat` + `gates` specs against the mock daemon, **2× full runs: 49 passed, 0 failed, 0 flaky** each.
- Retry-shaped run (`-g "revision"`, so the describe's earlier tests never run — the exact shape that was structurally red) `--repeat-each=3`: 6/6 passed.
- `pnpm --filter @qlan-ro/mainframe-ui typecheck` clean; `SessionRow` + `use-hover-card-wedge-guard` unit tests pass (6 new cases pin the gate, including the click-focus regression); `cargo check` / `cargo fmt --check` / `cargo test -p mainframe-adapter-mock` clean (2 new cases: a permission answer past the end is ignored, a `sendMessage` past the end still reports a desync).


doruchiulan:
Verified on Linux, the environment all three failures were exclusive to: full mock e2e suite via `gh workflow run e2e-mock.yml --ref fix/v2.0.0-e2e-gate-failures` — [run 31965278753](https://github.com/qlan-ro/mainframe/actions/runs/31965278753) is **399 passed, 0 failed, 0 flaky, 36 skipped** (6.4m). The v2.0.0 gate run was 396 passed / 2 failed / 1 flaky over the same 399. That run also builds the daemon from source, so it is the first one to exercise the mock-adapter change (the local macOS runs used `MF_E2E_SKIP_BUILD=1` against a pre-change binary; the Rust change is covered locally by its two `cargo test` cases).

PR #688, opened 2026-09-13
feat(chat): migrate desktop messaging to an ACP session façade

doruchiulan:
## What changed

Move daemon-to-desktop chat messaging onto an ACP v2 façade at `/acp/{profile}`, using a pinned protocol snapshot and negotiated Mainframe extensions.

- Add JSON-RPC initialization, prompt/cancel, stable transcript items, streaming patches, resume replay, and permission resolution across attached clients.
- Migrate the desktop chat controller, message projection, and permission UI to the façade.
- Adopt Claude partial-message streaming and preserve provider IDs through live output and history.
- Retire the legacy chat transcript, gate, queue, compaction, and usage frames; publish queue/compaction/reset state through the façade's extension surface.
- Include the design, implementation plan, protocol reference, golden fixtures, unit/integration tests, and E2E coverage.

## Why

The current protocol repeatedly sends growing messages and reconstructs chat state through several separate reconnect paths. The façade provides explicit lifecycle messages and one replay contract for desktop chat.

Implements Mainframe todo 350. The native vendor adapters remain in place. Project/storage interfaces and other daemon event domains retain their existing interfaces. Mobile migration is explicitly outside this PR and review scope.

## Testing done

- `git diff --check origin/main...HEAD` passed.
- `cargo test -p mainframe-acp --quiet`: 103 passed.
- `cargo test -p mainframe-server --test acp_ws_integration --quiet`: 11 passed.
- Shared-types and UI TypeScript checks passed.
- Focused desktop ACP client, accumulator, session-plane, and permission-answer tests: 56 passed across four files.
- All six GitHub checks passed.
- Adversarial review completed in two passes: changes needed for wrong-daemon routing, replay/live-update ordering, explicit transcript-clear handling, and retry-marker ordering. Isolated diagnostics reproduced the routing and transcript-clear regressions; findings are posted in the PR review. No source fixes were applied during review.
- Full local build and live E2E have not been rerun for this PR.

## Checklist

- [ ] `pnpm test` passes
- [ ] `pnpm build` passes (TypeScript compiles)
- [ ] No secrets in staged files
- [ ] File size limits respected (max 300 lines/file, 50 lines/function)

---

## Review-fixes lane rerun — 2026-09-14

Plan re-entered at "implement" on operator instruction (plan revised in place on this branch; the lane's reviewer loop was not re-run against the revision — this stage's APPROVED verdict is the operator's). Artifact: docs/plans/2026-09-14-pr-688-review-fixes-plan.md. Review verdict: APPROVED (1 round). QA: 8/8.

### Decisions
[plan] Lane re-entered at "implement" on operator instruction. The plan was revised in place on todo/350-wire-protocol-payload-grammar and the lane's reviewer loop was NOT re-run against it — this stage's APPROVED verdict is the operator's, not the reviewers'. Task groups re-extracted from the plan at HEAD.

[implement] T6 gap found by review: the plan required correcting stream.rs's module doc to state the new ordering fact once raw out-of-band frames joined the throttle FIFO, and to explain why GateResolved is the one exception (direct fan-out to the gate-holder, not the session). This was missed in the original T6 commit and added as a follow-up commit (5d70ab94) after the group's other 11 commits were already pushed.

[implement] T2: ChatManager::is_chat_working already existed as pub(super) in chat_manager/construct.rs; changed visibility to pub rather than adding a duplicate method (first attempt hit E0592 duplicate-definition).

[implement] T3: apply_gate_answer previously called the concrete ChatManager directly, not a port; added a GatePort trait in ports.rs so a failing test double (FlakyPort) could drive the failed-apply-leaves-gate-answerable path. Test now lives in dispatch/gate_answers/tests.rs after T10's file split.

[implement] T5: R1.2 (diff/publish atomicity) test pins behavior rather than observing a deterministic red (per plan decision 6, no deterministic red exists there); R2.9 (revision during snapshot-await is buffered) was genuinely red before the fix. SessionSlot::{Live, AwaitingSeed} plus FacadeHub::begin_resume is the buffering mechanism that closes both. hub.rs (356 lines) and hub/tests.rs (517 lines) are pre-existing over-cap files the plan assigns to a later task (T37), not newly pushed over cap by this group's edits.

[implement] T7: adding options/scope fields to ControlRequest/ControlResponse (no Default derive) broke ~23 construction sites across mainframe-adapter-claude, mainframe-adapter-codex, mainframe-chat and their tests; all fixed with options: None / scope: None (Codex real wiring is deferred to T19 per plan, commented accordingly) - no behavior change in those crates. Separately, implementing adapter-supplied options surfaced a real pre-existing correctness gap: Claude's default options unconditionally offered allow-always even with zero suggestions, contradicting spec decision 12/D4; fixed as a natural consequence of the same change, not deferred.

[implement] T8: the group's step-0 fixture is tests/support/facade.rs (an injectable-adapter FacadeServer harness) plus, for T10, tests/support/barrier_adapter.rs. The new session_detach test lives in its own acp_ws_session_detach.rs file to keep acp_ws_integration.rs under the 300-line cap; attach-on-send happens before dispatch_with_prompt, so the detach test didn't need a real ChatManager.

[implement] T9: initialize_required() reuses error_codes::RESOURCE_NOT_FOUND (-32002) per plan, exported from mainframe-acp so mainframe-server's session/resume gate reuses the exact same error object rather than a second definition. The _mainframe.dev/session_detach notification is intentionally NOT gated by the negotiated check (it's a notification with no reply either way). Found and fixed session_prompt_reaches_the_prompt_port in acp_ws_integration.rs as 'passing by coincidence' (same numeric error code, wrong underlying cause, because it never called initialize first) - recorded as a correctness fix, not just a compile fix, via a new connect_initialized() helper.

[implement] T10: built a full minimal Adapter/AdapterSession fake (BarrierAdapter) rather than reusing MockCliAdapter/ReplaySession, because ReplaySession::emit fires its delayed events from a detached spawned task and so never actually blocks spawn()'s own await - it could not produce a genuine, non-timing-dependent block for the test. Both connection.rs and dispatch.rs crossed 300 lines purely from this task's own additions (not pre-existing) and were decomposed: connection.rs's test module extracted to connection/tests.rs; dispatch.rs's gate-answer block extracted to dispatch/gate_answers.rs with its own test module.

[implement] T11: net::client_ip became unused in production after the fix (trust_proxy_client_ip is now used for the WS upgrade path) but was deliberately NOT removed - the file's PORT STATUS comment block documents 1:1 historical correspondence with the original TypeScript source, a documentation convention this group's other tasks treated as non-removable, not dead code. Flagging this as a judgment call for review rather than silently deleting or silently leaving it unexplained.

[implement] No foreign-file staging incidents occurred on any of the 12 commits in this group - git diff --cached --stat was checked before every commit and always listed only this group's own files.

[implement] Post-hoc downstream check (beyond the plan's literal T12 verify commands): cargo clippy --workspace --all-targets -D warnings is clean; pnpm --filter @qlan-ro/mainframe-ui typecheck is clean against the adapter.ts changes from T7. Grepping packages/ui and packages/e2e surfaced several files that assert or narrate the OLD unconditional allow-once/allow-always/reject-once gate contract and the pre-T2/T9/T10 session/resume and prompt-reply ordering assumptions - specifically packages/e2e/FLOW-MAP.md:115 ('the daemon offers allow-once / allow-always / reject-once on every gate'), packages/e2e/tests-tauri/gates.spec.ts and chat.spec.ts (assert chat-permission-option-allow-always is always visible), and packages/e2e/tests-tauri/facade-protocol.spec.ts / stress-matrix.spec.ts (session/resume redelivery and reply-ordering assertions). These are plausibly affected by T7's Claude-suggestions-gated allow-always and by T2's added trailing state_update frame on resume, but they are e2e specs requiring a daemon rebuild, owned by a different group, and out of this group's scope - reported here rather than fixed.

[implement] Prior work found: origin/main..HEAD already had T1-T11 (group 1) plus docs/plan commits landed before I started; I started clean from T13, no rework needed.

[implement] T13: new PartialOverlays type keyed by (chat_id, session_id) in a new event_handler/partial_overlay.rs module; on_exit's own-overlay clear moved above the existing session-id guard so a stale sink's exit cannot touch a newer session's overlay.

[implement] T16: edited mainframe-acp/src/stream.rs though the plan's Files list for T16 omitted it — the Change text explicitly required narrowing attach_retry_marker's carrier selection there, so I followed the Change text over the Files list per T16's own guidance precedent.

[implement] T17: ports.rs's queued_position-for-commands fix has no dedicated red test (would require a real ChatManager + mainframe-server-crate test); only the chat-crate TurnStarted-guard half is test-covered by the plan's own verify line. Implemented both fixes, documented the asymmetry in the commit message.

[implement] T19: no synthetic 'reject' option for requestUserInput (D4 says real choices only); reused the legacy updatedInput.answers shape (keyed by question text) rather than inventing a new meta key, so resolve() needed zero new branches for the allow path; ExitPlanMode's options split AllowOnce/RejectOnce by a 'no'-prefix label heuristic mirroring find_by_prefix; flipped ask_user_question_falls_back_to_empty_string from behavior:deny to behavior:allow since deny-with-nothing is now a genuine decline (a new test pins that shape).

[implement] T21 (R2.8, Claude): confirmed and fixed a real defect via git-stash red/green verification — a signature-only thinking entry claimed the API message id live but history's convert_assistant_entry drops that entry without claiming it, so the next content-bearing entry disagreed on id between live and history. Fixed with has_representable_content mirroring history's per-block keep rule.

[implement] T21 (Codex parity) + T22 (partial): built tests/live_vs_history_id_parity.rs against the existing collab-delegation-0.144.3.jsonl fixture. Fixed R3.17 (convert_thread_items silently dropped ThreadItem::DynamicToolCall via a catch-all _ => {}) by adding a reload arm that reuses thread_item_render's naming/input helpers (promoted to pub(crate)). Did NOT fix or pin R3.16 (needs a multi-turn fixture this crate lacks) or R3.18 (lives in rollout_reader.rs's raw-.jsonl-rescan path, a different subsystem than history_convert.rs, which the plan's own text pointed at incorrectly) — documented both as named, unverified gaps in the test file's module doc and the commit message rather than fabricating unverified #[ignore]'d tests.

[implement] Over-cap files touched but not decomposed (T37/group-4 owns the decomposition pass explicitly per the plan): chat_manager/send.rs (314), permission_handler.rs (393, was already 387 pre-existing), acp_ws/hub.rs (398), approval_handler.rs (428), message_parsing.rs (478, pre-existing at 458), tests/approval_handler.rs (405, pre-existing at 309). None of these were under 300 before my edits.

[implement] git diff --cached --stat was checked before every commit; only my own files were ever staged — no foreign paths from sibling groups appeared at any point.

[implement] Step 0 (mandated by the plan) extracted AcpSessionPlane's attach/reattach/resumeFromGap/guard/listener-wiring into a new packages/ui/src/features/chat/controller/acp-session-attachment.ts, keeping AcpSessionPlane's public API byte-identical so the existing 23-test suite was the oracle for the pure move.

[implement] T24: reattach() pre-resets the accumulator+settled cursor AND passes bypassGuard through to resume(), so a live update racing the resume() round-trip can't re-arm the empty-refresh guard against a genuine server wipe. New direct unit tests were added against AcpSessionAttachment (acp-session-attachment.test.ts) rather than only through the plane, since the guard/bypass distinction is cleaner to isolate at that layer; the plan's cited acp-session-plane.ts:206 refusal test was rewritten in place per the plan's explicit instruction.

[implement] T25: reconcile now feeds the matcher only the suffix of user messages since the last transcript.updated dispatch (AcpSessionPlane.newUserMessagesSinceLastDispatch, a high-water-mark getter with a side-effecting count update), not the whole accumulated history. resetAccumulator() also explicitly zeroes the count so a resume/reattach re-baselines to 0 — this combined with stable-id upsert idempotency means a replay that re-establishes the same history never emits a spurious reconcile batch, with no special resume-detection branch needed in the getter itself. local.message.failed now accepts an optional event-carried `pending` fallback (chat-actions.ts already has the original pending object in scope at both call sites) so a failed send is never silently dropped when the clientId was already removed from state.

[implement] T25 advisor guidance on attachment-only reconcile keying (cross-matching between two ATTACHMENT_KEY pendings) was read as already closed by the suffix-only feed (each echo lands as exactly one new item, so oldest-first matching is correct without a clientId-based key) — no separate change made; recorded per the advisor's explicit permission to skip it and note the reading.

[implement] T27: connect() now builds the RpcConnection locally and assigns this.connection only after the handshake AND version check pass, closing the attempt's own socket in a catch on any failure. handleClose(connection) takes the connection instance so a leaked failed-connect() socket closing late can't tear down a connection that has since replaced it.

[implement] T29: sendRequest() arms a 30s deadline (REQUEST_DEADLINE_MS) cleared on settle; write() on a dead socket rejects that specific pending entry (not just logs); close() rejects all pending synchronously rather than relying on the underlying WebSocket's asynchronous onclose event.

[implement] T30: RpcConnection gained respondError(); AcpNotificationRouter.handleRequest takes an injected onRequestError callback (matching the existing onHeartbeat injection pattern) rather than importing RpcConnection directly, keeping the router transport-agnostic. AcpFacadeClient wires it to `this.connection?.respondError(...)` so it always targets the current connection across reconnects.

[implement] T31: parseItemMeta's per-field fallback was extracted into a new packages/ui/src/features/chat/view-model/parse-item-meta.ts rather than inlined in convert-acp-item.ts, since convert-acp-item.ts was already at 287 lines and the addition would have breached the 300-line cap.

[implement] T26: getAcpFacadeClient keys on `${getActiveDaemon().id}:${profile}` read at resolve time; disposeDaemonSession() also hard-clears the registry via resetAcpFacadeClients() so a daemon switch that lands back on the SAME daemon id (a re-pair/token refresh) still gets a fresh client rather than reusing one that could be mid-teardown.

[implement] T34: pending user messages now carry an optional sendMeta (PromptSendMeta) computed before the pending is built in sendChatMessage, and retryChatMessage resends pending.sendMeta ?? {} instead of always {}. Deleted AcpFacadeClient.onClose/CloseListener (verified no production consumer via grep; only the client's own test used it, which was rewritten to assert client.connected instead). Added the full _mainframe.dev/resync path end-to-end (ResyncListener/handleResync in the notification router, onResync on AcpFacadeClient, wiring in AcpSessionAttachment calling reattach() without dispatching transcript.cleared) — the daemon-side production caller for this notification was already live from group 1/2's work (mainframe-acp::capabilities::resync_notification, hub.rs:364), only the client-side listener was missing.

[implement] T32: projectChatThreadMessages appends queued turns built from state.interactions.queued (sorted by ref.timestamp, FIFO), skipping refs whose messageId already appears in the confirmed transcript (server message ids). Verified via a Rust source read that QueuedMessageRef.messageId is literally 'the display message ID (from MessageCache)' (mainframe-types/src/chat.rs:337), i.e. the SAME id the encoder uses as the item/container id once a queued message is dequeued and becomes a real transcript entry — so the dedup-by-id check is sound. Deleted the `queued` branch from convert-acp-user.ts's coerceUserMeta per the plan (the projection now sets meta.queued=true itself); rewrote one existing convert-acp-item.test.ts case that still asserted the old wire-sourced `queued` field. sendChatMessage now reads sendPrompt's `{queued}` return value and immediately dispatches local.message.reconciled for its own pending when true, so a queued send doesn't render beside its own optimistic bubble.

[implement] T33 (the largest task): split AcpSessionAttachment's client BINDING (bindClient — sets a reference only, no wire traffic; prompt/cancel/reply work from a dormant chat since the daemon attaches-on-send server-side per dispatch.rs:63-71, verified by reading the Rust source) from its live SUBSCRIPTION (a new `subscribed` boolean gating listener-wiring + session/resume). Added detach() (sends _mainframe.dev/session_detach, unsubscribes, keeps client bound) and reactivate() (re-subscribes from the last settled item via a cursor resume on a switch-back, or falls back to attach()'s full {type:'start'} replay if never attached before) as new methods alongside the existing attach()/reattach()/resumeFromGap(), which were left semantically untouched to avoid breaking T24's guard test. AcpChatController gained setActive(active) (delegating to a new extracted ChatActivation helper) gating the plane's subscription the same way subscribeLive gates the side-band WS, both driven from the SAME opts.active effect in use-chat-thread-runtime.ts. ChatPlaneLoader.load() now only seeds REST config and binds the client (host.bindPlane); it ALSO checks a new host.isActive() callback and calls host.reactivatePlane(client) inline within the SAME deduped promise when active — this was a deliberate deviation from a naive two-separate-async-operations design, made specifically to avoid a microtask-ordering race between the mount effect's load() and the activation effect's setActive(true) (found via manual reasoning about promise-continuation scheduling order, not directly stated in the plan); recorded here since a reviewer re-deriving the timing would otherwise have to redo that analysis. Redefined loadState:'ready' to mean 'config seeded + client bound' (plus 'facade subscribed' only when active) rather than always implying a completed resume — justified because only the active thread ever renders, so an inactive thread's 'ready' without a live stream is harmless; this satisfies the plan's own suggestion to make that call. Extracted ChatActivation and ChatLiveSubscription into new files to bring acp-chat-controller.ts back under the 300-line cap after the T33 additions (it hit 334 lines pre-extraction). One pre-existing test (chat-thread-controller-local-load.test.ts's 'loads against the real id... resumeCalls' assertion) encoded the OLD always-attach-on-adopt behavior that D2 deliberately removes; rewritten into two tests pinning the new contract (inactive adopt -> no resume; active adopt -> resume), justified inline as a behavior change by design.

[implement] makeController() in acp-test-kit.ts now defaults to active:true (calling ctrl.setActive(true) synchronously at construction) to keep ~40 pre-existing controller tests (reconcile, cancel, send, retry, T25's tests) passing without per-test changes, since they all implicitly relied on the plane streaming once loaded; dormancy-specific tests pass {active:false} explicitly or construct AcpChatController directly and call setActive() themselves. This was the lower-churn of the two options the advisor offered and is called out in the file's module doc.

[implement] Adopted from prior sessions: groups 1-3 (T1-T34) were already committed on the branch at start; read and built on top without redoing any of it.

[implement] T35: the first attach-on-prompt test only proved attach ran before the prompt's OWN reply arrived -- with the no-ChatManager harness that reply comes back instantly either way, so a mutation that moved attach to AFTER dispatch resolves left it green. Added a second test using the existing BarrierAdapter fixture: the adapter's spawn blocks indefinitely, a resume for an unrelated chat on the same socket proves the prompt frame's own handle_inbound call already returned (FIFO frame processing), and a manually-pushed revision for the still-blocked chat must reach the socket regardless. Verified red against the late-attach mutation, restored, committed as a follow-up (packages/core-rs/crates/mainframe-server/tests/acp_ws_prompt_subscribes.rs).

[implement] T35: renamed event_mapper.rs's dynamic_tool_call_vendor_id_matches_the_item_id_history_reload_would_use to dynamic_tool_call_completion_uses_the_item_id_as_its_vendor_id, not the plan's literally-proposed '..._history_does_not_render_it' -- verified T22 already landed and fixed the history-drop defect with its own two-sided parity test in live_vs_history_id_parity.rs (dynamic_tool_call_reload_matches_the_live_tool_use_id_and_name, comment says '(fixed)'). The plan's proposed name was written before that fix shipped and would itself have been false; the new name states what the test actually asserts (only the live path).

[implement] T36: used z.toJSONSchema()-based recursive key-declaration walker rather than a schema.shape walker or a strictified-clone oracle (both plan-sanctioned alternatives). Two real bugs in the walker itself, found and fixed before it produced zero false positives on the real fixture set: (1) naive branch-matching on a plain z.union picked the first branch with ANY matching const property, misidentifying update.ts's three-variant state_update family and jsonrpc.ts's result/error split -- fixed with a scored branchFit (const-match count first, key-overlap count as tiebreaker); (2) tool-call.ts's rawInput: z.unknown().nullish() decomposes to anyOf[{}, {type:'null'}] with neither branch declaring `properties`, which the walker must treat as fully open (record/unknown), not flag every key underneath as undeclared.

[implement] T37 scope: measured every file >300 lines in the full origin/main...HEAD diff against its line count at a78c9f5b (the review-fix plan's stated branch head). In scope = the plan's 7 named files + any file <=300 at a78c9f5b and >300 now (a genuine new crossing from this review-fix plan's own tasks). That yielded 12 files, all fixed: encoder.rs, session_state.rs, hub.rs, encoder/tests.rs, session_state/tests.rs, hub/tests.rs, facade-protocol.spec.ts (plan-named); stream/tests.rs, partial_overlay_tests.rs, chat_manager/send.rs, partial_stream/tests.rs, project-messages.test.ts (new crossings). Additionally decomposed two functions clippy -W clippy::too_many_lines found still over 50 lines after the file splits and grown by this same review-fix plan's own T5/T6/T9/T10/T35 work: handle_acp_socket (acp_ws.rs) and handle_inbound (acp_ws/dispatch.rs).

[implement] T37 explicitly OUT of scope, left untouched, documented here per the advisor's guidance: dozens of files across the workspace that were already over 300 lines at a78c9f5b and merely grew further under earlier groups' edits -- same class as event_handler.rs's own explicit 'must not grow, need not shrink' carve-out in the plan's Constraints section. Largest: event_handler.rs (2304->2293, shrank), chat_manager/tests.rs (2081, untouched), chat_deps.rs (2041, untouched), adapter-claude/session.rs (1799, untouched), adapter-claude/events.rs (1650->1653), adapter-codex/event_mapper.rs (1103->1108, my own T35 rename touched it trivially), websocket.rs (707->729, T11 grew it), types/adapter.ts (675->695), convert-acp-item.test.ts (589->600), message_parsing.rs (458->478, T14 grew it), acp-session-plane.test.ts (454->455), approval_handler.rs src (399->428) and tests (309->405, T19 grew both), chat_surface_wiring.rs (374->414), permission_handler.rs (387->393), acp-client.test.ts (304->347, already over before any review-fix work).

[implement] Added packages/core-rs/clippy.toml with too-many-lines-threshold = 50 as a standing diagnostic (kept, not dropped after use) -- verified it does NOT affect the plan's actual CI command: `cargo clippy --all-targets -- -D warnings` (no explicit -W) stayed clean across the full workspace both before and after adding the file, because clippy::too_many_lines is in the pedantic/opt-in group and the threshold config only takes effect when the lint is explicitly enabled with -W. Running with -W clippy::too_many_lines (a manual diagnostic, not CI) surfaces roughly 90 pre-existing violations workspace-wide that predate this todo entirely and are out of scope.

[implement] T38: the spec's Decisions list ended at 27 (verified by reading the file directly, not trusting a stale line number); appended 28-34, all marked reversible (none constitute a structural pivot on the order of Decisions 1/2/3/24, which chose the facade direction, the frozen snapshot, and stable-id canonicalization). Documented _mainframe.dev/session_detach and _mainframe.dev/resync in API-REFERENCE.md's ACP Chat Facade section (grep-located via its heading, not a cited line number, since the plan's line 726 citation was from a78c9f5b) plus T10's reply-ordering note.

[implement] Verified at every commit (per instructions) that git diff --cached --stat listed only my own files before committing; no sibling-group files ever appeared staged -- no lint-staged interference incidents to report.

[implement] e2e: not run. This environment has no built Tauri daemon binary and no browser runtime; the repo's own memory notes 'batch e2e at the end' as a whole-series concern, not a per-lane one. Verified the one e2e file I split (facade-protocol.spec.ts -> facade-protocol.spec.ts + facade-protocol-streaming.spec.ts + helpers/tauri/facade-protocol-support.ts) preserves all 10 tests via `E2E_MODE=mock npx playwright test --list`, plus tsc --noEmit and eslint clean on all three files.

### Non-blocking review notes

[reviewer] `FacadeHub::reset_session` resurrects an attachment the client already dropped. `packages/core-rs/crates/mainframe-server/src/acp_ws/hub.rs:121-128` removes whatever slot exists and then inserts `SessionSlot::Live(stream)` unconditionally. If `_mainframe.dev/session_detach` (or `ChatEnded`) lands while `handle_resume`'s `dispatch_resume` await is in flight (`acp_ws/dispatch.rs:190-207`), `forget_chat` drops the slot and the completing resume re-creates it — the daemon then encodes and pushes that chat's updates to a connection whose client listeners are gone, until the next detach. The client side degrades safely (listeners removed; the stale `lastSettledItemId` is still a valid cursor on switch-back), so this is a dormancy/bandwidth regression against T33's acceptance rather than corruption. One-line fix: only insert when `sessions.contains_key(chat_id)` still holds (or track a detach generation), and skip `deliver` in that case.

[reviewer] Four files added by this branch exceed the 300-line cap T37's acceptance says none may: `packages/ui/src/features/chat/view-model/__tests__/convert-acp-item.test.ts` (600), `packages/ui/src/features/chat/controller/__tests__/acp-session-plane.test.ts` (455), `packages/core-rs/crates/mainframe-chat/src/chat_manager/tests/chat_surface_wiring.rs` (414), `packages/ui/src/lib/daemon/__tests__/acp-client.test.ts` (347). T37 split the files it named and the ones it grew, but these — added by the earlier facade passes — were left. `event_handler.rs` at 2293 is under its 2304 baseline and is not a violation.

[reviewer] None of the six end-to-end scenarios the plan's Verification step 7 mandates were authored. The e2e diff for this group is `facade-protocol.spec.ts` split into `facade-protocol-streaming.spec.ts` plus a helper (T37) — no test covers a retry-interrupted partial stream leaving no ghost bubble, plan-mode clear staying cleared, a queued prompt staying at the tail through its dequeue, Codex always-allow not re-prompting, a daemon switch routing the next prompt, or a mid-turn reconnect resuming with the tail still streaming. The first and fourth are precisely the blockers whose unit tests pass at the seam's daemon end only (see must_fix #1).

[reviewer] `FacadeConnection::prompt_locks` grows without bound: `facade_conn.rs:71-78` inserts a per-session `tokio::sync::Mutex` on first prompt, and `forget_chat` (`facade_conn.rs:125-129`) clears `sessions` and `pending_gates` but not `prompt_locks`. A long-lived connection accumulates one entry per chat ever prompted. Bounded by chat count, so a slow leak — clear it in `forget_chat`.

[reviewer] `initialize` is not mandatory for attaching. `dispatch.rs:63-72` calls `ctx.facade_hub.attach()` for a `session/prompt` before `dispatch_with_prompt` applies the `negotiated` refusal, and `dispatch_fallback` (`dispatch.rs:89-93`) does the same for `session/cancel`. The prompt/cancel themselves are correctly refused, but the connection is left attached and starts receiving that chat's `session/update` frames — in a grammar a version-mismatched peer just declined. Not a security hole (the WS upgrade token still gates access), but it undercuts spec decision 32's "initialize before any session method". Move both `attach` calls behind the `is_negotiated()` check.

[quality] `facade_hub.attach` runs before the negotiation gate, so an un-negotiated (or version-mismatched) peer still gets a live stream. `packages/core-rs/crates/mainframe-server/src/acp_ws/dispatch.rs:67-69` attaches on `session/prompt` and `:89-91` attaches on `session/cancel`, both BEFORE `dispatch_with_prompt` applies the `negotiated` check (`mainframe-acp/src/connection.rs:64-66`). The method itself is refused, but the connection now holds a `SessionSlot::Live`, so `hub.rs:161-167`/`handlers.rs:53-63` encode and fan out every `session/update` for that chat to it. That contradicts spec decision 32 ('initialize is mandatory before any session method') and defeats the encode-only-when-listening guard. Fix: wrap both attach calls in `if connection.is_negotiated()`.

[quality] `FacadeConnection.prompt_locks` is never pruned. It is populated per distinct session in `session_prompt_lock` (`packages/core-rs/crates/mainframe-server/src/acp_ws/facade_conn.rs:71-78`) but `forget_chat` (`:125-129`) clears only `sessions` and `pending_gates`. Every chat a connection ever prompted leaves a permanent `Arc<tokio::sync::Mutex>` entry for the connection's lifetime, surviving both `ChatEnded` and `_mainframe.dev/session_detach`. Fix: add `self.prompt_locks.lock()...remove(chat_id);` to `forget_chat`.

[quality] Four hub handlers swallow serialization failures silently, violating the no-silent-catch rule, and repeat the same four-line shape. `packages/core-rs/crates/mainframe-server/src/acp_ws/hub/handlers.rs:115, 122, 131, 142` all use a bare `if let Ok(payload) = serde_json::to_string(&note)` with no else arm, while the sibling `handle_gate_raised` (`:71-74`) correctly `warn!`s on the same failure. Fix: extract one `fn push_notification<T: Serialize>(&self, chat_id: &str, note: &T, kind: &str)` that serializes, warns on failure, and calls `push_raw_to_attached` — collapsing queue_changed/transcript_cleared/resync/compaction to one line each and making the logging uniform.

[quality] Session-id extraction is duplicated and the hot path clones a whole request to get one string. `packages/core-rs/crates/mainframe-server/src/acp_ws/dispatch.rs:67` builds `InboundFrame::Request(request.clone())` purely so `prompt_session_id` (`:143-157`) can read `params.sessionId`, and then `spawn_prompt` (`:118-123`) re-implements the same extraction on the original request. Fix: extract the session id once before the branch, pass it into `spawn_prompt`, and drop the clone.

[quality] `reply_is_ok` decides whether the handshake succeeded by re-parsing the daemon's own just-serialized reply string back into `serde_json::Value` and probing for a `result` key (`packages/core-rs/crates/mainframe-server/src/acp_ws/dispatch.rs:92-104`). That is a string round-trip across a boundary we own on both sides, and it silently reads 'not ok' for any future reply shape change. Fix: have `dispatch_with_prompt` return a typed outcome (or mark `negotiated` inside the `initialize` arm in `connection.rs:110`, which is the only place that knows the answer) so the shell never has to re-parse.

[quality] `packages/ui/src/lib/daemon/acp-notification-router.ts` is seven near-identical copies of the same three-part pattern (a `Set` field at `:45-51`, an `onX` registrar at `:58-91`, and a `handleX` at `:132-172`), differing only in schema, listener signature, and which params it forwards. This fix range added an eighth copy (`resync`, T20/T34) rather than collapsing it — adding one `_mainframe.dev/*` notification currently costs four edits in three places. Fix: a single `Map<method, Set<listener>>` plus a table of `{ method, schema, project }` entries; `handleNotification` becomes a table lookup and ~110 lines collapse to ~30.

[quality] Raw out-of-band frames still bypass the throttle FIFO during the `AwaitingSeed` window. `packages/core-rs/crates/mainframe-server/src/acp_ws/hub.rs:224-237`'s `_ =>` arm sends directly for a connection whose slot is `AwaitingSeed`, and the comment openly calls this 'matching the pre-T6 behavior for that narrow window'. Because `handle_resume` calls `begin_resume` before awaiting the snapshot (`dispatch.rs:189-192`), a `transcript_cleared` (or a gate raise) that lands during the await reaches the client BEFORE the replay `reset_session` delivers — i.e. the client wipes and then receives the pre-wipe snapshot. Display revisions are already buffered for exactly this reason (`SessionSlot::AwaitingSeed { latest }`); raws are not. Fix: buffer raw frames alongside `latest` in `AwaitingSeed` and drain them after `deliver` in `reset_session`, or state explicitly in the spec why the window is acceptable.

[quality] `handle_retry` (`packages/core-rs/crates/mainframe-server/src/acp_ws/hub/handlers.rs:101-107`) drives the fan-out helper `for_each_attached_session` with a closure that mutates stream state and returns `Vec::new()` — using a send-updates-to-every-connection helper purely as a for-each with an inert return. It reads as a bug at a glance. Fix: add a small `for_each_attached_stream` that takes `impl FnMut(&mut SessionStream)` and returns nothing, and use it here.


doruchiulan (commented):
Review verdict: **changes needed** after two review passes and verification against the code. Reviewed `origin/main...a78c9f5b24feb8ae53a995e85bce8aef208b087a`; mobile migration is excluded. Three high-priority and one medium-priority findings remain open. No implementation changes were made during review.

1. **P1 — Daemon switching leaves chat traffic connected to the previous daemon.** [acp-clients.ts:12](https://github.com/qlan-ro/mainframe/blob/a78c9f5b24feb8ae53a995e85bce8aef208b087a/packages/ui/src/lib/daemon/acp-clients.ts#L12)

   The global registry is keyed only by adapter profile, and `disposeDaemonSession()` never disconnects or clears it. After switching from daemon A to B, `ensureConnected()` returns the old resolved initialization promise. REST requests use B, but ACP resume/prompt still use A. A reproduction using the production registry, client, and teardown with fake sockets sent a prompt for `chat-b` to `ws://daemon-a:31415/acp/claude?token=a`. Clear the registry during daemon teardown and bind cached clients to daemon identity.

2. **P1 — A live chunk can overtake replay and corrupt the displayed transcript.** [hub.rs:152](https://github.com/qlan-ro/mainframe/blob/a78c9f5b24feb8ae53a995e85bce8aef208b087a/packages/core-rs/crates/mainframe-server/src/acp_ws/hub.rs#L152)

   `for_each_attached_session()` computes updates under the session lock but releases it before enqueueing them. `reset_session()` uses that lock to seed and publish replay. A reachable interleaving is: live diff computes `lo` against `Hel`; resume seeds and queues the full `Hello`; the previously computed `lo` is queued afterward. The client displays `Hellolo` while the server believes it holds `Hello`, so later suffixes preserve the corruption. Keep diff computation and publication in the same ordering critical section as replay, or serialize them through one session actor. This finding is established by code-path analysis, not a concurrent runtime reproduction.

3. **P1 — Explicit transcript clears leave old items in the accumulator.** [acp-session-plane.ts:104](https://github.com/qlan-ro/mainframe/blob/a78c9f5b24feb8ae53a995e85bce8aef208b087a/packages/ui/src/features/chat/controller/acp-session-plane.ts#L104)

   `transcript_cleared` empties the reducer's visible messages but retains the accumulator. If the subsequent resume returns `itemCount: 0`, the empty-refresh guard rejects the reset. The next incoming message projects the retained items again. Reproduced with an explicit clear: expected message IDs `[new]`, received `[old, new]`. Explicit clears must invalidate the accumulator, timestamps, and cursor immediately and bypass the guard intended for ambiguous empty history reads.

4. **P2 — Provider retry metadata is attached after the invalidation update has already passed.** [event_handler.rs:1404](https://github.com/qlan-ro/mainframe/blob/a78c9f5b24feb8ae53a995e85bce8aef208b087a/packages/core-rs/crates/mainframe-chat/src/event_handler.rs#L1404)

   `on_api_retry()` clears the partial overlay and emits its display revision before emitting `ChatSurfaceEvent::Retry`. The content-clearing upsert therefore reaches `SessionStream` before `pending_retry` is set; the marker attaches to a later update instead. Existing stream/hub tests feed these events in the opposite order. Record the retry before emitting the clearing revision and test the production event order. This is a protocol-contract defect; no current desktop consumer reads the retry marker.

Validation: 103 Rust ACP tests, 11 server ACP integration tests, 56 focused desktop tests, both TypeScript checks, and `git diff --check` passed. All six GitHub checks passed. Full local build and live E2E were not rerun. The isolated diagnostic tests reproduced findings 1 and 3; they live outside the repository at `/tmp/mainframe-acp-pr-Mmuw29/review-repros.test.ts` and run with `pnpm --filter @qlan-ro/mainframe-ui exec vitest run --config /tmp/mainframe-acp-pr-Mmuw29/vitest.config.mjs`.

Triage note: stale gates after cancellation or a resolution missed during disconnection are real existing defects, but the relevant base-branch paths already exhibit them. The removed three-second permission tracker only verified a user's submitted reply; it did not clear cancelled gates. Those were excluded from this PR's regression findings after checking `origin/main`.

doruchiulan (commented):
**Verdict: changes needed.** Adversarial review, two iterations, every finding verified against `a78c9f5b` in the worktree. This complements the earlier review on this PR: its findings 1, 2, and 4 are new and I confirmed all three in code; its finding 3 is the same defect as my blocker 2 below, reported once here.

Merge consequence to reconfirm: the mobile client still reads `display.message.*` and `permission.requested` and still sends `message.send`. Spec decision 24 accepts this, but merging breaks mobile until its own repo lands the façade migration.

### Blockers

1. **An aborted partial stream leaves a permanent blank assistant bubble.** `packages/ui/src/features/chat/view-model/acp-item-accumulator.ts:156-172`. The daemon clears a vanished item with an empty-content upsert and forgets it (`mainframe-acp/src/session_state.rs:80-95`). The client keeps it: `applyUpsert` stores `content: []`, nothing removes the id from `order`, and `ensureNonEmpty` renders an empty text part. Every Claude `api_error` retry re-streams under a new provider message id, so the wire carries "clear A, create B" and a blank bubble sits above the real answer for the session's life. Fix: drop the item on an empty-content clear.

2. **Plan-mode "clear context" deterministically resurrects the wiped transcript.** `acp-session-plane.ts:125-128, 199-205`. `plan_mode_handler.rs:80-87` clears the Claude session id before `clear_messages()`, so the post-wipe snapshot is always empty. `reattach()` never resets the accumulator, and the empty-refresh guard refuses the `itemCount: 0` replay. The daemon reseeds its diff state to empty regardless, so nothing can ever clear the stale items. Same defect as finding 3 of the earlier review. Fix: `reattach()` resets the accumulator; a server-initiated wipe must bypass the guard.

3. **Every visited chat attaches to the façade and nothing detaches.** `use-chat-thread-runtime.ts:86-89`. The comment in `acp-chat-controller.ts:136` says attach-at-load is deliberate, but the spec and plan never mention dormancy and the cost is recorded nowhere. No detach exists in the protocol. After visiting N chats the daemon encodes, diffs, and pushes every background agent's every revision, and the hub's "encode only when someone is listening" guard never fires again. Either record the tradeoff or gate attachment on the active thread and add a detach.

### Should fix before merge

4. **Resume replays items but no turn state.** `mainframe-acp/src/resume.rs:56-68`. The tail message gets `status: running` only from run state (`project-messages.ts:71-79`), and that status is what makes the smooth reveal stream. After a mid-turn reconnect the tail no longer streams.

5. **A failed permission answer is silently discarded and wedges the turn.** `acp_ws/dispatch.rs:192, 203-211`. `remove_gate` runs before `claim_gate`, so on the error path a re-sent answer matches no pending gate. The client already dispatched the optimistic resolve. Gate vanishes, CLI stays blocked.

6. **The heartbeat watchdog fires once and never re-arms.** `acp-heartbeat-watchdog.ts:36-39`. Only a received heartbeat re-arms. A wedged-but-open socket gets one resume attempt and then silence; the catch comment in `acp-session-plane.ts:216` promising a later retry is false for that case.

7. **`/acp/{profile}` widens the read surface under the spoofable proxy rule.** `net.rs:20-31` vs `41-57`. Socket upgrade auth uses the leftmost-hop `client_ip` rule (the open audit item); HTTP uses `trust_proxy_client_ip`. On main the full display list travelled over the socket only incidentally during a live turn. `session/resume` makes any chat's stored history readable on demand under the weaker rule. Switching socket auth to the stricter rule closes both.

### Lower severity

8. Live and history may claim the provider message id on different entries (`assistant_event.rs:182-189` vs `history_converters.rs:346-397`) for a signature-only thinking block. Whether it surfaces as differing item ids is unconfirmed. `tests/live_vs_history_id_parity.rs` has no such case; add it first.
9. The resume snapshot is read outside the lock that seeds it (`dispatch.rs:95` vs `:109`). Lost revision self-heals unless it was the turn's last.
10. Golden fixtures cannot catch optional-field drift: all TS ACP schemas are `.loose()`, so a Rust-required field unmodelled in TS is invisible.
11. Gate and side-band notifications bypass the throttle FIFO (`hub.rs:228, 261, 267, 277`); a permission request can precede its tool call by up to 100 ms. `stream.rs:11` overstates the guarantee.
12. `acp-session-plane.ts:109` `.catch(() => undefined)` is a silent catch on blocker 2's path with no `/* expected */`. `AcpFacadeClient.onClose` has no consumer.
13. Size rules, which the PR checklist leaves unchecked: 11 new files over 300 lines (`encoder.rs` 448, `session_state.rs` 321, `hub.rs` 301, `facade-protocol.spec.ts` 398, plus 7 test files); `event_handler.rs` grew 2103 to 2304 in a branch whose commit `134e0313` claims to decompose it; `encode_content` ~140 lines and `on_chat_surface_event` ~108.

### Verified clean

Socket auth is a byte-identical extraction. No await under any std mutex. No non-mobile consumer of any retired frame survives. Command-name regex intact in `prompt.rs`. Accumulator reset is microtask-ordered ahead of the next socket message. Rejected as pre-existing: unbounded outbound channel, inline awaits in the socket loop. Rejected: the gate registry's single claimed slot (per-connection pending map is the real queue); dropping `permission.requested` from the connection-global set (`chat.updated` fires on the same raise).

### What ran

Types and UI builds pass locally in the worktree. Rust fmt, clippy, and tests green in CI on the head commit, plus the six JS checks. `git diff --check` clean. Nobody has run live E2E on this branch.

doruchiulan (commented):
**Third-pass addendum.** Five zones neither earlier review had touched: Claude partial streaming, client projection, the JSON-RPC transport, the Codex adapter path, and queue semantics. Two reviewers ran in parallel (Codex backend and clean-context Claude), two iterations per zone, every zone no-ship. Everything below is new, verified against `a78c9f5b` in the worktree, and excludes the 17 findings already on this PR.

### New blockers

1. **A dequeued prompt's reorder cannot be expressed on the façade; the transcript is wrong live and rearranges on reload.** `event_handler.rs:478` (`strip_queued_and_move`, called from `:1096` and the orphan sweep at `:875`); `mainframe-acp/src/session_state.rs:28,45`; `acp-item-accumulator.ts:77,136`. The dequeue path removes the prompt and pushes it to the tail. `SessionState` is a per-id `HashMap` and `diff()` compares per id, so a pure permutation emits nothing but a meta patch; the client's `order` list is append-only first-seen. On main `display_emitter.rs` detected `order_changed` and sent a full `DisplayMessagesSet`; that repair is gone. Scenario: turn 1 streams, user queues "B", agent emits more output, turn ends, CLI acks B. Daemon order is `[U1, A1, T1, A2, B]`; client renders `[U1, A1, B, T1, A2]`. Reconnect replays daemon order and the transcript visibly rearranges. Two clients that resumed at different moments disagree forever.

2. **"Always allow" is offered on every Codex gate and does nothing (regression).** `mainframe-acp/src/gates.rs:41-46,131`; `mainframe-adapter-codex/src/approval_handler.rs:168,235`; `PermissionGate.tsx`. On main the button was gated on `hasSuggestions`, and Codex builds every `ControlRequest` with `suggestions: Vec::new()`, so it was hidden. The façade's fixed three-option list offers it unconditionally; `parse_answer` collapses allow-always into `Allow` with no permissions, and `resolve` never emits Codex's `acceptForSession`. The user clicks Always allow on `cargo test` and the same gate blocks again next turn.

3. **Full-transcript re-reconcile over-clears optimistic sends (regression).** `chat-reconcile.ts:77-99` driven from `acp-chat-controller.ts:271-275`. On main the live path fed one `message.added` to the matcher; now `dispatchFromPlane` re-runs it against the entire accumulated user-message list on every `transcript.updated`, i.e. every streaming chunk. Type "continue" at turn 3 having typed it at turn 1: the history duplicate satisfies the new pending before its own echo arrives. If that send then fails, `local.message.failed` hits `if (!current) return state` and the message is lost with no failure indicator. Attachment-only sends all share one key.

4. **Every failed `initialize` leaks the losing connection.** `acp-client.ts:128-149`; `connection.rs:77-96`. `this.connection = connection` precedes `open()` and the handshake; the daemon answers a version mismatch with an error and keeps the socket open by design. An initialize error, a schema throw, or the version check all reject `connect()` while leaving the socket open, `this.connection` pointing at an un-negotiated connection, no watchdog, no reconnect. Each retry overwrites the field and the daemon keeps every socket with its own heartbeat ticker. `requireConnection()` still succeeds, so `prompt()` sends on a connection that never negotiated. A leaked socket closing later runs `handleClose` and tears down the healthy one.

5. **Codex `AskUserQuestion` gets an empty answer, and Reject is identical to Allow, for any plain ACP answer.** `gates.rs:33`; `approval_handler.rs:213-232,297-308`. The fixed option list cannot carry Codex's real choices; a plain `{optionId}` answer yields `updated_input: None`, `extract_answer_from_updated_input` returns `""`, and the `requestUserInput` arm never reads `behavior`. The desktop is unaffected because it sends the rich `_mainframe.dev` answer, but a generic ACP client (the stated reason for the façade) and any rich answer that fails `rich_answer`'s id validation both send Codex a blank answer for Allow and Reject alike.

### Should fix

6. **`session/cancel` is starved behind a slow prompt dispatch.** `acp_ws.rs:84-97`. `handle_inbound` is awaited inline in the `select!` arm; while it runs the loop drains no output, sends no heartbeat, reads no frames. The loop shape is inherited from `/`, but cancel moved from the REST `/interrupt` route onto this same socket, so a cold-chat start on chat A now blocks Stop for every chat on the profile. This upgrades the head-of-line item the first review rejected as pre-existing.
7. **A schema-rejected `session/request_permission` leaves the daemon's gate outstanding forever.** `acp-notification-router.ts:104-109`; `acp-rpc-connection.ts:109-111`. `handleRequest` returns early with no reply; `respond()` can only emit a result, never a JSON-RPC error; `gate_registry.rs` has no TTL. Outer-params skew produces no UI and no answer, and the turn hangs until the CLI dies. Distinct from the inner `controlRequest` leniency in e6efd3e9.
8. **No request deadline; a request issued after `close()` hangs forever.** `acp-rpc-connection.ts:91-102,113-119`. `sendRequest` registers pending then `write()` drops the frame on a null socket; `close()` never rejects pending. `prompt()` awaits forever and run state never leaves sending.
9. **Command-tag stripping is not streaming-safe.** `event_handler.rs:1414` vs `message_parsing.rs:338-379`. Both strippers require the close tag, so every partial before it passes the raw `<mainframe-command…>` wrapper to the UI, then the closing emission collapses the text. Streamed text shrinks. `partial_overlay_tests.rs:197` only feeds an already-closed tag.
10. **All-or-nothing `ItemMetaSchema.safeParse` deletes subagent transcripts silently.** `convert-acp-item.ts:43-46,269-282`. One wrong-typed known key drops `containerId`, `parentToolCallId`, `groupId`, and `subagent` together; children already diverted into `children` are never visited by any container.
11. **Cache eviction permanently diverges an attached client.** `acp-item-accumulator.ts:93-98`; `session_state.rs:58-76`; `message_cache.rs:8,50-51`. Past 2000 messages the cache evicts from the front, `clear_vanished` excludes tool calls, and no resume re-seed fires. The client has no removal operation at all.
12. **Daemon slash commands bypass the queue but report `queued_position`, and fire `TurnStarted` mid-turn.** `send.rs:235`; `ports.rs:57`. The bypass itself is pre-existing; the spurious running state and the false queued position are new.
13. **Attach-on-prompt has no test.** `dispatch.rs:52-56,63-77`. The deleted 190-line `ws_first_send_fanout.rs` pinned implicit subscription on send and exactly-once delivery; `acp_ws_integration.rs:134` runs with no `ChatManager` and asserts only an error code. Move `attach` below `dispatch_with_prompt` and the suite stays green while every prompting client goes deaf.
14. **`on_compact` is the one sink method the stable-id pass skipped.** `event_handler.rs:1214` uses `transient` (nanoid) while this PR adds `transient_with_id` beside it; history uses the entry id. Compaction pills always differ live vs reload, for Claude and Codex.
15. **No Codex live-vs-history equality test.** `tests/event_mapper.rs:519,591` pin vendor ids that `group_messages` discards before the wire. No Codex counterpart to `stable_id_characterization_tests.rs`. This is why 14, 16, and 17 ship undetected.

### Lower

16. Sub-agent card resolved by the child's own `turn/completed` re-attaches to the wrong turn on Codex reload (`turn_lifecycle.rs:44` vs `history_collab_resolve.rs:226`); the asymmetry is pre-existing, criterion 4 now asserts an equality it never held.
17. `dynamicToolCall` renders live and vanishes on reload (`thread_item_render.rs:48,101` vs `history_convert.rs:158`); pre-existing, but the new test `dynamic_tool_call_vendor_id_matches_the_item_id_history_reload_would_use` is named after an invariant history never satisfies.
18. Codex sub-agent history mints `rollout-{n}` ids for nested tool calls where live uses the app-server item id (`history_load.rs:51-62`). Pre-existing.
19. A superseded session's dangling partial overlay is never cleared: `on_exit`'s session-id guard at `event_handler.rs:1128` sits above `take_partial_overlay()` at `:1137`. Race, not deterministic; the fix needs session ownership on the overlay, not moving the clear.
20. Version-probe failure caches a silent permanent downgrade with no log line (`partial_stream.rs:220-240`), and the probe is awaited inline in `spawn()` adding up to 5 s to a cold start.
21. `initialize` is advisory: a version-mismatched peer keeps its socket and can prompt, resume, cancel (`dispatch.rs:42-58`).
22. The criterion-5 acceptance `_meta` has no consumer (`sendChatMessage` discards `queued`), so nothing would catch `queued_position` regressing; and `retryChatMessage` re-sends with `{}` meta, so retrying a failed `/command` ships the literal slash text.

### Rejected

Unbounded outbound channel (pre-existing on `/`, same as pass one). Late `on_message` after exit (pre-existing). `todoList` asymmetry (both paths deliberately emit nothing). `ParentIdSink` not forwarding partials (Codex never calls them). Cancelling a queued prompt via `session/cancel` (traced: it goes to REST `cancelQueuedMessage` and never touches the running turn). Codex parent/child item-id collision (unprovable from the repo). A throw inside `convertAcpItems` (no producer-reachable input found).

### Verified clean

`queue_state` closes every resume including empty. Two clients prompting while running both learn of the enqueue. Archive/end mid-queue drops gate and session state. Request-id spaces cannot collide. Malformed frames are dropped on both sides without crashing. Capability skew degrades to the fallback heartbeat. Concurrent subagent/parent overlays are keyed by chat and parent deltas are dropped. Item order within a turn only grows at the tail. `patchField` null-vs-omit matches the fixtures.

One approach note, not a defect: queue mutation (cancel, edit) stayed on REST while everything else moved to the façade. A generic ACP client can enqueue and observe but has no protocol-level way to cancel or edit a queued prompt.

doruchiulan (commented):
**Review-fix rounds complete at `26c4c5f0`.** Everything the three reviews and the lane's 13 non-blocking notes raised is closed, plus what six further one-shot review rounds on the fix range surfaced. CI (typecheck, unit suites, changeset, Rust fmt/clippy/test) is green on this head; the full mock E2E suite passes locally at this head: 412 passed, 0 failed, 0 flaky, 35 pre-existing skips.

### What landed after the lane's PR stage (68 commits)

- **All 13 lane notes.** Detach mid-resume no longer resurrects an attachment; `initialize` gates attach as well as the method; prompt locks pruned only when unheld; hub handlers log serialization failures; session id extracted once; typed handshake outcome; raw frames buffered in the await window and drained after the replay; `handle_retry` uses a for-each; `client_ip` deleted and the port-status comment corrected; over-cap files split; the notification router is table-driven.
- **E2E scenarios.** Four of the six mandated scenarios are authored and green (retry leaves no ghost bubble, plan-mode clear stays cleared, queued prompt stays at the tail, mid-turn reconnect keeps streaming). The mock adapter gained queue acknowledgement and a plan-mode handler to make two of them drivable. Codex always-allow and daemon switch are recorded in `COVERAGE-GAPS.md` with their unit-test pins, since mock mode cannot drive them.
- **Regressions the rounds caught and fixed.** The clear frame is now explicit (`content: []` plus `_meta: null`), so skill-loaded and compaction pills render again. Cancel and resume moved behind the same per-session lock as the prompt, with the lock taken on the socket loop under `unconstrained` so a spent coop budget cannot lose a queue place. Lifecycle events raised during a resume await are buffered as stream ops and drained after the replay. Gate raises are registered and delivered in one pass. A gate resolved during the await, or since the snapshot, is never redelivered. A panicked prompt, cancel, or resume answers its request with `-32603`; a failed resume asks for one resync per failure streak, and the client backs off between full replays with a single in-flight guard for wipes and resyncs. Gate options are adapter-supplied with a rich answer standing on its request-id match; empty or blank Codex labels are never offered and never answered blank. The reconcile matcher keys on stable item ids so a replay cannot clear a fresh pending.
- **Spec decisions 28 to 35** record the new protocol behaviour; the changeset describes the user-visible changes.

### Known, not in this PR

- The mobile client still speaks the retired dialect (ratified, decision 24).
- `packages/e2e` has 95 pre-existing TypeScript 6 type errors on `main` (no Node types in its tsconfig); CI does not run that typecheck.
- Pre-existing over-cap files touched by the range grew by a few lines each (`event_handler.rs`, `websocket.rs`, `message_parsing.rs`, the two `adapter` type files); `handle_request` in the Codex intake is ~150 lines and was moved verbatim. Both are separate decomposition work.

PR #700, opened 2026-09-21
fix(ui): let Enter send after an unrecognized trigger token (#353)

doruchiulan:
## Summary

Typing an unrecognized `/` or `@` token (e.g. `/compact`) in the composer swallowed the Enter key: the only way to send was Escape then a click on Send. The trigger-field key handler now falls through Enter/arrow keys/Tab to normal composer behavior when the current token has no matching entries, so Enter sends immediately. Escape's focus-park behavior now gates on the real keyboard event instead of a stale "armed" flag, so focus correctly stays in the composer after dismissing an unmatched token. The same underlying fix makes Enter insert a newline (instead of doing nothing) in automation trigger text fields.

## Related work

Todo #353.

Plan: `docs/plans/2026-09-21-todo-353-unrecognized-command-enter-plan.md`

## Changes

- `use-trigger-field.ts`: Enter/ArrowUp/ArrowDown/Tab fall through when the armed token has zero matching entries; Escape/Backspace keep their own guards since an empty list there is legitimate (e.g. drilled into an empty category).
- `Composer.tsx` / `ComposerTriggers.tsx` / `trigger-field-aria-context.tsx`: added a sibling `armed` context (not folded into the ARIA props object, since that gets spread onto the textarea DOM node) so the Escape focus-park branch reacts to the real armed state rather than stale `aria-expanded`.
- See the [Implementation notes](https://github.com/qlan-ro/mainframe/pull/700#issuecomment-5756657975) comment for the async `@`-mention risk and other tradeoffs surfaced in review.

## Validation

- `pnpm --filter @qlan-ro/mainframe-ui test` on all touched and adjacent trigger-engine specs: 120+ tests green, including new regression pins for the unmatched-token chip, Escape-then-Enter submit, and the automations-field newline case.
- `pnpm exec tsc --noEmit -p .` in `packages/ui`: clean.
- Prettier on all changed files: clean (package has no separate lint script).
- Live QA: 5/7 scenarios passed against the built app; see Visual evidence.

## Risks and rollout

- While an async `@`-mention search is in flight, the entry list is momentarily empty, so Enter now sends the draft instead of the previous silent no-op. Scoped out of this fix's brief and accepted as a known tradeoff; mitigation would be treating an in-flight fetch as still-armed.
- Tab-after-an-unmatched-token now moves focus out of the composer (previously a silent no-op); intended per the fix but worth a UX gut-check.
- Full-suite run had one unrelated flake (`AutomationsView.test.tsx`, a lazy-import timeout) that passes in isolation and is untouched by this diff.

## Visual evidence

Live QA passed 5/7 scenarios against commit `884101ff`. Screenshots exist locally but automated upload to this PR is unavailable in this environment (no `gh pr comment --attach` support and no browser automation for the GitHub comment editor). See the Visual evidence status in the run notes; publishing is blocked pending manual upload.

## Review checklist

- [ ] Visual evidence is attached for visible changes, or the reason it is not needed is stated.


doruchiulan:
<details><summary>Implementation notes</summary>

- The "armed trigger" signal is threaded to the composer through a sibling context (`trigger-field-aria-context.tsx`), not a key on the ARIA props object. That object is spread directly onto the textarea DOM node, so a non-ARIA key would leak as a bogus attribute; and the plugin's own return value can't carry it either, since the composer's `onKeyDown` runs before assistant-ui's plugin loop.
- Escape stays consumed and default-prevented while a token is armed, matching prior behavior (document-level Escape listeners still don't fire on that keypress). Only the *focus-park* decision changed, from `aria-expanded !== true` to `!armed`; `armed` is a superset of `expanded`, so open-menu handling is unchanged.
- Async `@`-mention sources: while a fuzzy search is in flight, the entry list is empty, so Enter now sends the draft instead of the previous silent no-op. This is a real behavior change on a path the original brief scoped out; flagged as an accepted risk above rather than fixed here, since the adapter has no in-flight/loading signal to gate on yet.
- Review found the original guard comment redundant/miswritten and a test assertion that couldn't fail on its own; both were tightened in the final commits (`e65c86be`, `884101ff`) so Escape's disarm is asserted on an observable, not just a same-both-paths `false` return.

</details>

doruchiulan:
## Visual evidence

Tested SHA: `884101ff5bc50e8ab9c63d1b0f0e10bd0564c81b` (matches this PR's head).
Target: browser (renderer + daemon), isolated data dir, ports 31787/5539.

| Scenario | Result | Observation |
|---|---|---|
| S1 — unmatched `/token`, Enter alone | PASS | Typed `/zzz-nope`; suggestions stayed closed; **Enter alone sent it** (no Escape, no click) — the composer cleared and the turn landed with a command chip, agent replying "Unrecognized command: `/zzz-nope`." |
| S2 — `/` with a matching entry | PASS | Typed `/`; popover opened with 1 match (`launch-config`); ArrowDown+Enter **inserted** the entry into the composer and did **not** submit. |
| S3 — unmatched `/token`, Escape then Enter | PASS | Typed an unmatched token, pressed Escape: popover gone and **focus stayed in the composer** (previously it jumped to the transcript, forcing a click on Send). A following Enter then sent the message normally. |
| S4 — empty composer, Escape (regression) | PASS | With no trigger armed, Escape still parks focus on the transcript viewport, unchanged from before this fix. |

Screenshots:

S1 — unmatched token sent by Enter alone, rendered as a command chip

S3 — after Escape, focus ring stays in the composer with the token text intact

Recorded clip — walks S1 then S3 live in the running app (~15s):

https://github.com/user-attachments/assets/925ecc76-b24f-4c75-9a31-6bc9712e0050

All four scenarios ran against the app built from the PR's head commit; build identity was verified by inspecting the served (esbuild-transformed) `Composer.tsx` and `use-trigger-field.ts` sources over `/@fs/...` before testing.

PR #702, opened 2026-09-21
feat(ui): render transcript file paths as file-open links (#355)

doruchiulan:
## Summary

Transcript markdown links that point at a file (an absolute path, a project-relative path, or a `file://` URI, optionally suffixed with `:line` or `:line:col`) now render as file references instead of web links. Clicking one opens the file in the workspace via the existing open-file surface intent and jumps to the given line/column. Its context menu offers **Open file** plus **Copy absolute/relative path** instead of Copy link/Open link. A `file://` link, previously stripped to a dead anchor by the markdown sanitiser, now opens correctly. Regular `http(s)` links, the localhost tunnel chip, and smart-action chips are unchanged. The message-level path context menu (right-click on a tool-card file-path pill) also gained an **Open file** item alongside its existing copy actions.

A new pure classifier, `parseFileHref` (`packages/ui/src/lib/files/file-href.ts`), is shared by the markdown URL sanitiser and the link dispatcher so both agree on what counts as a file link. See [Implementation notes](https://github.com/qlan-ro/mainframe/pull/702#issuecomment-5757695743) for design rationale.

## Related work

Todo #355. Plan: `docs/plans/2026-09-21-transcript-file-path-links-plan.md`.

## Changes

- `parseFileHref` classifies a URL as a file reference (path + optional line/character); `urlTransform` relaxes the sanitiser for `file:` only when `key === 'href'`, so other URL-valued attributes (e.g. `img src`) are unaffected.
- `SmartLink` branches to a new `FileRefLink` component whenever `parseFileHref` matches, before any other link handling.
- `useOpenFile` now optionally accepts a `{ line, character }` position, additive to its existing single-argument call sites.
- **Known gap, needs a decision before merge:** per the plan's own open question, a failed file open still surfaces only as the workspace tab's existing "File not found or unreadable" state (no toast), because `emitSurfaceIntent` returns `void` and a toast at that layer would also fire for unrelated openers (file tree, spotlight, review panel, find-in-path). If a toast is required, the fallback is `mfToast.error` in `viewer-router.tsx`/`EditorTab.tsx`, at the cost of it firing for every other file-open failure path too.
- **Out of scope, flagged for follow-up:** `packages/ui/src/features/editor/MarkdownPreview.tsx` renders markdown through the same widened `urlTransform` but uses its own anchor (not `SmartLink`), which sends `file://` hrefs to `host.shell.openExternal` with no `.catch`. This surface isn't part of the five markdown surfaces this todo covers, so it was left untouched; a reviewer should confirm that's acceptable or file a fast follow-up.

## Validation

- `pnpm exec vitest run src/lib/files src/features/chat/parts src/features/chat/smart-actions src/features/chat/messages` — 30 files / 378 tests passed.
- `pnpm --filter @qlan-ro/mainframe-ui typecheck` — clean.
- `pnpm exec eslint` clean on all touched files.
- Live QA: 5/6 scenarios passed (absolute/relative/`file://` link rendering, workspace jump-to-line, distinct context menus). The one open item is the toast gap noted above.

## Risks and rollout

- Two known minor false positives in the classifier (documented trade-offs, not fixed here): a bare hostname-shaped string like `example.com` (no path) and a schemeless `//host/path` both classify as file references; neither was a working link before this change.
- No functional/rollback risk: the change only affects link classification and rendering in markdown surfaces already listed above; no data or API changes.

## Visual evidence

Live QA captured screenshots locally (transcript file-reference styling, workspace jump-to-line, and both context menus), but no GitHub-authenticated upload path was available in this run (`gh pr comment` has no attachment flag, and no browser upload session was available) — evidence is not published. See the linked QA report for the local captures.

## Review checklist

- [ ] Visual evidence is attached for visible changes, or the reason it is not needed is stated. (Capture exists locally; upload blocked — see Visual evidence.)


doruchiulan:
<details><summary>Implementation notes</summary>

- `toFileRef` (existing helper) is a normaliser, not a predicate — it treats an already-relative-looking string like `https://example.com` as a valid `FileRef`. That's why this change adds a separate pure `parseFileHref` classifier rather than reusing `toFileRef` for the is-this-a-file decision; `urlTransform` and `SmartLink` both call the new classifier so the sanitiser and the renderer can't drift on what counts as a file link.
- The `file:` relaxation in `urlTransform` is scoped to `key === 'href'` specifically because react-markdown calls `urlTransform` for every URL-valued attribute, including `img src` — an unscoped relaxation would have let live `file://` image sources load in the webview.
- Hrefs arrive percent-encoded (mdast-util-to-hast applies `normalizeUri`), so `parseFileHref` decodes non-`file://` targets in a try/catch; `file://` targets are left encoded since the existing `toFileRef` already handles decoding them.
- `FileRefLink` lives in `features/chat/parts/`, not `smart-actions/`, because that folder's convention gates every override there on the smart-actions provider, and this link must work without it.
- `link-with-preview.tsx` was deliberately left untouched — todo #341 ("Open in Mainframe" row on http links) is scheduled to land in that file, and the file-link branch lives in `SmartLink` plus the new `FileRefLink` component instead, to avoid a merge collision.

</details>

doruchiulan:
## Visual evidence

Tested at `c6ff47a01fecdb37de3de17c5a04c0069b46ace5` (matches current PR head). 5/6 live QA scenarios passed; the one gap — S6, missing toast on a failed file open — is also documented in the PR description's Known gap note.

| Scenario | Result | Caption |
|---|---|---|
| S1/S5 | PASS | File paths in the transcript (absolute, project-relative, `file://`) render as file-reference links with the file-code glyph, distinct from a real web link in the same message; a click jumps the workspace editor to the target line/column. |
| S2 | PASS | Right-clicking a file-reference link shows its dedicated context menu — **Open file**, **Copy absolute path**, **Copy relative path** — instead of the web link's Copy/Open link items. |
| S3/S6 | FAIL (AC gap) | Clicking a file reference to a nonexistent path opens a new workspace tab showing the viewer's in-tab "File not found" state; no toast fires, which falls short of the brief's "surfaces a toast" acceptance criterion. |

### S1 — transcript file-reference links
S1 - transcript file-reference links

### S2 — file-reference context menu
S2 - file-reference context menu

### S3 — missing file, no toast
S3 - missing file opens in-tab error state, no toast

PR #703, opened 2026-09-21
Fix new-session draft targeting and layout inheritance (#354)

doruchiulan:
## Summary

Starting a new session used to inherit whatever Workspace panels the previous session had open, and toggling those panels while composing a draft silently overwrote that session's remembered layout — so switching back to it, or sending the draft's first message, could leave the wrong arrangement on screen. Separately, "New" from the sidebar, the session-tab strip, and ⌘N always dropped the user on a "Choose a project" step even when the active project (or the session's own project) made the target obvious, and the welcome screen's project picker didn't reflect recent activity.

A `__LOCALID_*` draft is now a first-class layout key: it gets its own arrangement (chat-only by default), no longer clobbers the originating session's layout, and hands its arrangement to the real chat once the first message lands. All four "new session" entry points now resolve a target project (active filter, else the current session's project) up front and route through one shared hook; the welcome picker orders projects most-recently-active first.

## Related work

Todo #354.

## Changes

- `layout-sessions.ts` adds `dropSession`/`adoptSession` (kill-before-remove disposal on drop, source-wins on adopt) so a draft's Workspace state lives and dies independently of the session it was opened from.
- `use-start-new-session.ts` is the one hook all four entry points (`SessionsNewButton`, `SidebarActions`, `SessionTabs`, ⌘N) now call to resolve the target project and seed the draft; `use-open-draft.ts` and `use-new-chat-hotkey-handler.ts` are deleted as redundant.
- `resolve-new-session-project.ts` implements the resolution rule (active filter, else active session's project, else none); `WelcomeState`'s project picker sorts by recent activity.

## Validation

- `pnpm --filter @qlan-ro/mainframe-ui typecheck` and `eslint` clean on all touched files.
- Full UI vitest suite green (712 files / 6990 tests), including new/updated coverage for the resolver, the shared hook, the router's layout-follow behavior, and `WelcomeState`'s picker ordering.
- live-qa: 8/8 scenarios passed against the built app (tested commit `48f12791`), including draft-opens-chat-only and picker-ordering.

## Risks and rollout

- Two edge cases identified in review are not yet fixed: (1) clicking "+" while already on a projectless draft can point the draft's own discard target at itself, stranding the user on the reset draft instead of returning them to the prior session; (2) picking/changing a project from the welcome screen resets the draft's layout entry, so a Workspace opened before picking a project is lost. Both are pre-existing-shaped regressions of the new "+"-everywhere path, scoped to the discard/reassign flows rather than the core open/handoff path this PR targets.
- A non-active draft's layout entry can still be pruned by a thread-list reload that lands while the user is on a different session (the persisted-session GC only exempts the *active* draft). Low likelihood in practice — a configured draft is reset on navigate-away already — but not proven unreachable.
- No production code depends on the new `getPendingDraftProject()` export yet; it's kept as the imperative escape hatch the plan called for.

## Visual evidence

Not published: `gh pr comment` has no attachment flag in this environment and no authenticated browser session is available here to use the GitHub web uploader. Evidence exists locally from live-qa (2 screenshots, tested commit `48f12791`) but is not yet attached to this PR — see the lane record for capture details.

## Review checklist

- [ ] Visual evidence is attached for visible changes, or the reason it is not needed is stated.


doruchiulan:
## Visual evidence

Recovered browser QA evidence from tested commit `48f12791b94d8bca20184c31722f7a6c33fad899` (light theme). Current head `2f5a2ede` adds only changeset wording; the captured UI code is unchanged. This publication did not rerun QA.

**S1 — New draft opens chat-only with its project already selected.** The prior session's Workspace is not inherited.

S1: new draft is chat-only and scoped to the current project

**S7 — Welcome project picker follows recent activity.** The recorded assertion confirmed the same ordering as the sidebar; the newer QA project A appears above project B.

S7: project picker ordered by recent activity

The run recorded **8/8 planned scenarios passing**. It separately reproduced a timing window: issuing another New action immediately after a first-send handoff or discard can temporarily lose the inherited project. Zero-session tail ordering and the steady-state no-project fallback were not directly exercised. These limits remain relevant despite the passing scenario count.

